PR #53, opened 2018-07-26
Redacts x-ms-copy-source header.

jiacfan:
1. Redacts x-ms-copy-source header.
2. Optimize comments for highlevel.


jiacfan:
Close as sending PR from forked branch is not supported temporarily.

PR #54, opened 2018-07-27
Redacts x-ms-copy-source header.

jiacfan:
Redacts x-ms-copy-source header. Optimize comments for highlevel.


jiacfan:
Close as sending from forked branch is not supported temporarily.

PR #67, opened 2018-09-04
Fix UploadStreamToBlockBlob not reporting incoming stream errors

### LINKED ISSUE #52: UploadStreamToBlockBlob does not return error correctly
`UploadStreamToBlockBlob` which calls `uploadStream` is not returning error correctly to user. If reader is wrapped up with error, we were expecting this call to fail the transfer, instead currently below code simply copies stream to buffer irrespective of error type.

https://github.com/Azure/azure-storage-blob-go/blob/f0a732ea9441c73511367d5861da6bda664ffabd/2017-07-29/azblob/highlevel.go#L492

Once we break out of upload process here https://github.com/Azure/azure-storage-blob-go/blob/eaae161d9d5e07363f04ddb19d84d57efc66d1a1/2018-03-28/azblob/highlevel.go#L501
it simply masks occurred error and returns success to caller
https://github.com/Azure/azure-storage-blob-go/blob/eaae161d9d5e07363f04ddb19d84d57efc66d1a1/2018-03-28/azblob/highlevel.go#L509

I believe correct behavior here should fail the transfer and return error to user, this used be to be the behavior in older SDK [azure-sdk-for-go CreateBlockBlobFromReader](https://github.com/Azure/azure-sdk-for-go/blob/51228ee60b238eaa5a9d53adc82ebd8b321bca4f/storage/blockblob.go#L125)

Please let me know if you need any more details on this.

ncw:
Before this change, when using UploadStreamToBlockBlob, if the input
stream returned an error UploadStreamToBlockBlob would not return an
error and create a blob.

This is incorrect - when there is an error in the input stream
UploadStreamToBlockBlob should return an error and the blob should not
be created.

This fix makes sure the error is reported correctly and the blob is
not created in the input stream.

Fixes #52

----

I'm presenting this as a first iteration.

It doesn't have tests, though I have tested it fixes the specific problem I'm having with rclone.

If you think this is going in the right direction I'll write tests if I can figure out the test framework.

This aborts the blob without committing the block list.  This means there are uncommitted blocks hanging around for some time. These will be garbage collected after a week which seems acceptable to me, since this is a low probability event.  You may feel differently though!

Note also a fix to turn an unchecked type assertion into a checked type assertion - without this there is no way to return an error from `uploadStream` as I couldn't figure out how to make a `CommonResponse`.

### DIFF
--- a/2018-03-28/azblob/highlevel.go
+++ b/2018-03-28/azblob/highlevel.go
@@ -372,7 +372,8 @@ func UploadStreamToBlockBlob(ctx context.Context, reader io.Reader, blockBlobURL
 	result, err := uploadStream(ctx, reader,
 		UploadStreamOptions{BufferSize: o.BufferSize, MaxBuffers: o.MaxBuffers},
 		&uploadStreamToBlockBlobOptions{b: blockBlobURL, o: o, blockIDPrefix: newUUID()})
-	return result.(CommonResponse), err
+	commonResponse, _ := result.(CommonResponse)
+	return commonResponse, err
 }
 
 type uploadStreamToBlockBlobOptions struct {
@@ -471,6 +472,7 @@ func uploadStream(ctx context.Context, reader io.Reader, o UploadStreamOptions, 
 
 	// This goroutine grabs a buffer, reads from the stream into the buffer,
 	// and inserts the buffer into the outgoing channel to be uploaded
+	var readErr error
 	for c := uint32(0); true; c++ { // Iterate once per chunk
 		var buffer []byte
 		if numBuffers < o.MaxBuffers {
@@ -487,22 +489,33 @@ func uploadStream(ctx context.Context, reader io.Reader, o UploadStreamOptions, 
 			// We are at max buffers, block until we get to reuse one
 			buffer = <-incoming
 		}
-		n, err := io.ReadFull(reader, buffer)
-		if err != nil {
+		var n int
+		n, readErr = io.ReadFull(reader, buffer)
+		if readErr != nil {
+			if readErr != io.EOF && readErr != io.ErrUnexpectedEOF {
+				break
+			}
 			buffer = buffer[:n] // Make slice match the # of read bytes
 		}
+
 		if len(buffer) > 0 {
 			// Buffer not empty, upload it
 			wg.Add(1) // We're posting a buffer to be sent
 			outgoing <- OutgoingMsg{chunkNum: c, buffer: buffer}
 		}
-		if err != nil { // The reader is done, no more outgoing buffers
+		if readErr != nil { // The reader is done, no more outgoing buffers
+			readErr = nil
 			break
 		}
 	}
 	// NOTE: Don't close the incoming channel because the outgoing goroutines post buffers into it when they are done
 	close(outgoing) // Make all the outgoing goroutines terminate when this channel is empty
 	wg.Wait()       // Wait for all pending outgoing messages to complete
+	// Report the error from io.ReadFull if there is one
+	if readErr != nil {
+		return nil, readErr
+	}
 	// After all blocks uploaded, commit them to the blob & return the result
-	return t.end(ctx)
+	result, err := t.end(ctx)
+	return result, err
 }


ncw:
Not really sure what happened to the Travis tests - the errors didn't seem related to the code I changed - Help!

zezha-msft:
Hi @ncw, thanks for submitting the PR! We'll review it shortly.

Don't worry about the test failures, they are caused by the lack of test credentials when PRs come from outside remotes. Basically, Travis is protecting our test account credentials. 😄

2018-03-28/azblob/highlevel.go
```diff
@@ -487,22 +489,33 @@ func uploadStream(ctx context.Context, reader io.Reader, o UploadStreamOptions,
...
 	}
 	// NOTE: Don't close the incoming channel because the outgoing goroutines post buffers into it when they are done
 	close(outgoing) // Make all the outgoing goroutines terminate when this channel is empty
 	wg.Wait()       // Wait for all pending outgoing messages to complete
+	// Report the error from io.ReadFull if there is one
+	if readErr != nil {
+		return nil, readErr
+	}
 	// After all blocks uploaded, commit them to the blob & return the result
-	return t.end(ctx)
```
zezha-msft: this doesn't need to be changed, right?

2018-03-28/azblob/highlevel.go
```diff
@@ -372,7 +372,8 @@ func UploadStreamToBlockBlob(ctx context.Context, reader io.Reader, blockBlobURL
 	result, err := uploadStream(ctx, reader,
 		UploadStreamOptions{BufferSize: o.BufferSize, MaxBuffers: o.MaxBuffers},
 		&uploadStreamToBlockBlobOptions{b: blockBlobURL, o: o, blockIDPrefix: newUUID()})
-	return result.(CommonResponse), err
+	commonResponse, _ := result.(CommonResponse)
```
zezha-msft: Could you please rebase on top on the dev branch? A bit of code has changed here.

2018-03-28/azblob/highlevel.go
```diff
@@ -487,22 +489,33 @@ func uploadStream(ctx context.Context, reader io.Reader, o UploadStreamOptions,
 			// We are at max buffers, block until we get to reuse one
 			buffer = <-incoming
 		}
-		n, err := io.ReadFull(reader, buffer)
-		if err != nil {
+		var n int
+		n, readErr = io.ReadFull(reader, buffer)
+		if readErr != nil {
+			if readErr != io.EOF && readErr != io.ErrUnexpectedEOF {
```
zezha-msft: why io.ErrUnexpectedEOF?

2018-03-28/azblob/highlevel.go
ncw: OK

2018-03-28/azblob/highlevel.go
ncw: From the docs

```
func ReadFull(r Reader, buf []byte) (n int, err error)
    ReadFull reads exactly len(buf) bytes from r into buf. It returns the number
    of bytes copied and an error if fewer bytes were read. The error is EOF only
    if no bytes were read. If an EOF happens after reading some but not all the
    bytes, ReadFull returns ErrUnexpectedEOF. On return, n == len(buf) if and
    only if err == nil.
```

Since we aren't expecting the input buffer to be a multiple of `len(buffer)` in size then `is.ErrUnexpectedEOF` is an expected successful termination as well as `io.EOF`.

ncw:
I took at look at the dev branch and I would say this problem has been fixed there already. 

However the dev branch appears to have changed the API so I couldn't actually try it in my test code

```
# github.com/ncw/rclone/backend/azureblob
./azureblob.go:242:46: multiple-value azblob.NewSharedKeyCredential() in single-value context
./azureblob.go:573:17: undefined: azblob.ListContainersResponse
./azureblob.go:755:53: cannot use options (type azblob.BlobAccessConditions) as type azblob.ModifiedAccessConditions in argument to dstBlobURL.StartCopyFromURL
./azureblob.go:1132:37: not enough arguments in call to blockBlobURL.StageBlock
	have (context.Context, string, readSeeker, azblob.LeaseAccessConditions)
	want (context.Context, string, io.ReadSeeker, azblob.LeaseAccessConditions, []byte)
```

Was that intentional? I don't see a note in the README that the api is subject to change?

Is the plan that the dev branch will become the master branch at some point?  Or does the fix need backporting from the dev branch?  If the latter I'm happy to turn this PR into a backport of the fix from the dev branch.

zezha-msft:
Hi @ncw, you are right! It indeed looks like @jiacfan has already fixed this [issue](https://github.com/Azure/azure-storage-blob-go/commit/c7a2d66432994b955143d6a80f80f482c4784b09). I should have verified before asking for a PR, I'm very sorry! 😢 

As for the API change that you noticed, it was an intended fix. The dev branch, which contains a variety of bugs fixes and improvements, is going to be released soon to the master branch.

I apologize again for wasting your time, and really appreciate you taking time to submit the PR. Thank you!

ncw:
No worries, it wasn't time wasted, it was time invested in getting to know the SDK source code better :-)

Thanks

Nick

zezha-msft:
Thank you @ncw! It's heart-warming to hear that. Please don't hesitate if you see anything that needs improvement in the SDK, we'd love to have your contribution. 😃

PR #70, opened 2018-09-10
Adding recommendations for ways to remove panics.

marstr:
**This will not actually ever make its way to the upstream.**

These are a list of recommendations of how to refactor away the panics that have been identified in the Azure-Storage-Blob-Go repository.


zezha-msft:
Hi @marstr, could you please share which repo is this being pushed to? It would be nice for us to follow this discussion.

marstr:
You can see the discussion in [my fork of this repo](https://github.com/marstr/azure-storage-blob-go). We'll share the finalized document with you in a gist soon though. :)

Issue #72, opened 2018-09-22
OpenBSD and NetBSD build issue


kpfaulkner:
Hi

I'm cross compiling one of my apps that I've just ported to this new SDK (from azure-sdk-for-go). I'm using gox to trigger a bunch of cross platform builds.

Most platforms are fine, but OpenBSD and NetBSD fail.

gox -ldflags "-X main.Version=0.4.0"
Number of parallel builds: 3

-->      darwin/386: azurecopy/azurecopycommand
-->    darwin/amd64: azurecopy/azurecopycommand
-->     linux/amd64: azurecopy/azurecopycommand
-->       linux/arm: azurecopy/azurecopycommand
-->       linux/386: azurecopy/azurecopycommand
-->     freebsd/386: azurecopy/azurecopycommand
-->   freebsd/amd64: azurecopy/azurecopycommand
-->     openbsd/386: azurecopy/azurecopycommand
-->   openbsd/amd64: azurecopy/azurecopycommand
-->     windows/386: azurecopy/azurecopycommand
-->   windows/amd64: azurecopy/azurecopycommand
-->     freebsd/arm: azurecopy/azurecopycommand
-->      netbsd/386: azurecopy/azurecopycommand
-->    netbsd/amd64: azurecopy/azurecopycommand
-->      netbsd/arm: azurecopy/azurecopycommand

5 errors occurred:
--> openbsd/386 error: exit status 2
Stderr: # github.com/azure/azure-storage-blob-go/2016-05-31/azblob
..\..\github.com\azure\azure-storage-blob-go\2016-05-31\azblob\highlevel.go:165:7: undefined: mmf
..\..\github.com\azure\azure-storage-blob-go\2016-05-31\azblob\highlevel.go:167:12: undefined: newMMF

--> openbsd/amd64 error: exit status 2
Stderr: # github.com/azure/azure-storage-blob-go/2016-05-31/azblob
..\..\github.com\azure\azure-storage-blob-go\2016-05-31\azblob\highlevel.go:165:7: undefined: mmf
..\..\github.com\azure\azure-storage-blob-go\2016-05-31\azblob\highlevel.go:167:12: undefined: newMMF

--> netbsd/386 error: exit status 2
Stderr: # github.com/azure/azure-storage-blob-go/2016-05-31/azblob
..\..\github.com\azure\azure-storage-blob-go\2016-05-31\azblob\highlevel.go:165:7: undefined: mmf
[...]

PR #73, opened 2018-10-01
Remove panics

### LINKED ISSUE #59: Please don't panic on DecodeString
I was surprised to see all of the implementations of `NewSharedKeyCredential` panic when decoding the account key.

https://github.com/Azure/azure-storage-blob-go/blob/197d1c0aea1b9eedbbaee0a1a32bf81e879bde80/2018-03-28/azblob/zc_credential_shared_key.go#L23

I would expect the error to be returned, rather than a panic.

@JeffreyRichter, there are over 100 panic calls in this library. Is this intended? I haven't found an instance of a panic call which wouldn't be better handled by returning an error. Thoughts?

zezha-msft:
Fix #59 and #72.

### DIFF
--- a/azblob/url_container.go
+++ b/azblob/url_container.go
@@ -3,6 +3,7 @@ package azblob
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 
@@ -86,7 +87,7 @@ func (c ContainerURL) Create(ctx context.Context, metadata Metadata, publicAcces
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/delete-container.
 func (c ContainerURL) Delete(ctx context.Context, ac ContainerAccessConditions) (*ContainerDeleteResponse, error) {
 	if ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		sanityCheckFailed("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
+		return nil, errors.New("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
 	}
 
 	ifModifiedSince, ifUnmodifiedSince, _, _ := ac.ModifiedAccessConditions.pointers()
@@ -106,7 +107,7 @@ func (c ContainerURL) GetProperties(ctx context.Context, ac LeaseAccessCondition
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/set-container-metadata.
 func (c ContainerURL) SetMetadata(ctx context.Context, metadata Metadata, ac ContainerAccessConditions) (*ContainerSetMetadataResponse, error) {
 	if !ac.IfUnmodifiedSince.IsZero() || ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		sanityCheckFailed("the IfUnmodifiedSince, IfMatch, and IfNoneMatch must have their default values because they are ignored by the blob service")
+		return nil, errors.New("the IfUnmodifiedSince, IfMatch, and IfNoneMatch must have their default values because they are ignored by the blob service")
 	}
 	ifModifiedSince, _, _, _ := ac.ModifiedAccessConditions.pointers()
 	return c.client.SetMetadata(ctx, nil, ac.LeaseAccessConditions.pointers(), metadata, ifModifiedSince, nil)
@@ -178,7 +179,7 @@ func (p *AccessPolicyPermission) Parse(s string) error {
 func (c ContainerURL) SetAccessPolicy(ctx context.Context, accessType PublicAccessType, si []SignedIdentifier,
 	ac ContainerAccessConditions) (*ContainerSetAccessPolicyResponse, error) {
 	if ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		sanityCheckFailed("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
+		return nil, errors.New("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
 	}
 	ifModifiedSince, ifUnmodifiedSince, _, _ := ac.ModifiedAccessConditions.pointers()
 	return c.client.SetAccessPolicy(ctx, si, nil, ac.LeaseAccessConditions.pointers(),
@@ -238,7 +239,7 @@ func (c ContainerURL) ListBlobsFlatSegment(ctx context.Context, marker Marker, o
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/list-blobs.
 func (c ContainerURL) ListBlobsHierarchySegment(ctx context.Context, marker Marker, delimiter string, o ListBlobsSegmentOptions) (*ListBlobsHierarchySegmentResponse, error) {
 	if o.Details.Snapshots {
-		sanityCheckFailed("snapshots are not supported in this listing operation")
+		return nil, errors.New("snapshots are not supported in this listing operation")
 	}
 	prefix, include, maxResults := o.pointers()
 	return c.client.ListBlobHierarchySegment(ctx, delimiter, prefix, marker.val, maxResults, include, nil, nil)
--- a/azblob/url_container.go
+++ b/azblob/url_container.go
@@ -16,9 +16,6 @@ type ContainerURL struct {
 
 // NewContainerURL creates a ContainerURL object using the specified URL and request policy pipeline.
 func NewContainerURL(url url.URL, p pipeline.Pipeline) ContainerURL {
-	if p == nil {
-		panic("p can't be nil")
-	}
 	client := newContainerClient(url, p)
 	return ContainerURL{client: client}
 }
@@ -89,7 +86,7 @@ func (c ContainerURL) Create(ctx context.Context, metadata Metadata, publicAcces
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/delete-container.
 func (c ContainerURL) Delete(ctx context.Context, ac ContainerAccessConditions) (*ContainerDeleteResponse, error) {
 	if ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		panic("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
+		sanityCheckFailed("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
 	}
 
 	ifModifiedSince, ifUnmodifiedSince, _, _ := ac.ModifiedAccessConditions.pointers()
@@ -109,7 +106,7 @@ func (c ContainerURL) GetProperties(ctx context.Context, ac LeaseAccessCondition
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/set-container-metadata.
 func (c ContainerURL) SetMetadata(ctx context.Context, metadata Metadata, ac ContainerAccessConditions) (*ContainerSetMetadataResponse, error) {
 	if !ac.IfUnmodifiedSince.IsZero() || ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		panic("the IfUnmodifiedSince, IfMatch, and IfNoneMatch must have their default values because they are ignored by the blob service")
+		sanityCheckFailed("the IfUnmodifiedSince, IfMatch, and IfNoneMatch must have their default values because they are ignored by the blob service")
 	}
 	ifModifiedSince, _, _, _ := ac.ModifiedAccessConditions.pointers()
 	return c.client.SetMetadata(ctx, nil, ac.LeaseAccessConditions.pointers(), metadata, ifModifiedSince, nil)
@@ -181,7 +178,7 @@ func (p *AccessPolicyPermission) Parse(s string) error {
 func (c ContainerURL) SetAccessPolicy(ctx context.Context, accessType PublicAccessType, si []SignedIdentifier,
 	ac ContainerAccessConditions) (*ContainerSetAccessPolicyResponse, error) {
 	if ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		panic("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
+		sanityCheckFailed("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
 	}
 	ifModifiedSince, ifUnmodifiedSince, _, _ := ac.ModifiedAccessConditions.pointers()
 	return c.client.SetAccessPolicy(ctx, si, nil, ac.LeaseAccessConditions.pointers(),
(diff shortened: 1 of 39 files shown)


azblob/zc_credential_shared_key.go
```diff
@@ -164,7 +164,7 @@ func (f *SharedKeyCredential) buildCanonicalizedResource(u *url.URL) string {
 	// params is a map[string][]string; param name is key; params values is []string
 	params, err := url.ParseQuery(u.RawQuery) // Returns URL decoded values
 	if err != nil {
-		panic(err)
+		sanityCheckFailed("parsing query parameters must succeed, otherwise there might be serious problems in the SDK/generated code")
```
devigned: `sanityCheckFailed` seems like a shallow alias to an underlying panic. Is that accurate?

azblob/zc_uuid.go
```diff
@@ -21,10 +21,7 @@ type uuid [16]byte
 func newUUID() (u uuid) {
 	u = uuid{}
 	// Set all bits to randomly (or pseudo-randomly) chosen values.
-	_, err := rand.Read(u[:])
-	if err != nil {
-		panic("ran.Read failed")
-	}
+	rand.Read(u[:])
```
devigned: Are we ok with the case read fails?

azblob/zc_sas_account.go
```diff
@@ -22,17 +23,17 @@ type AccountSASSignatureValues struct {
...
-		panic("Account SAS is missing at least one of these: ExpiryTime, Permissions, Service, or ResourceType")
+		return SASQueryParameters{}, errors.New("account SAS is missing at least one of these: ExpiryTime, Permissions, Service, or ResourceType")
 	}
 	if v.Version == "" {
 		v.Version = SASVersion
 	}
 	perms := &AccountSASPermissions{}
 	if err := perms.Parse(v.Permissions); err != nil {
-		panic(err)
+		return SASQueryParameters{}, err
```
devigned: Parsing the permissions here causes a panic. If these permissions could originate from supplied input, perhaps, this should return an error rather than panic.

azblob/url_container.go
```diff
@@ -109,7 +106,7 @@ func (c ContainerURL) GetProperties(ctx context.Context, ac LeaseAccessCondition
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/set-container-metadata.
 func (c ContainerURL) SetMetadata(ctx context.Context, metadata Metadata, ac ContainerAccessConditions) (*ContainerSetMetadataResponse, error) {
 	if !ac.IfUnmodifiedSince.IsZero() || ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		panic("the IfUnmodifiedSince, IfMatch, and IfNoneMatch must have their default values because they are ignored by the blob service")
+		sanityCheckFailed("the IfUnmodifiedSince, IfMatch, and IfNoneMatch must have their default values because they are ignored by the blob service")
```
devigned: This function already has an error on the return signature. Why not return an error here rather than a panic?

azblob/url_container.go
```diff
@@ -181,7 +178,7 @@ func (p *AccessPolicyPermission) Parse(s string) error {
 func (c ContainerURL) SetAccessPolicy(ctx context.Context, accessType PublicAccessType, si []SignedIdentifier,
 	ac ContainerAccessConditions) (*ContainerSetAccessPolicyResponse, error) {
 	if ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		panic("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
+		sanityCheckFailed("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
```
devigned: This function already has an error on the return signature. Why not return an error here rather than a panic?

azblob/url_container.go
```diff
@@ -241,7 +238,7 @@ func (c ContainerURL) ListBlobsFlatSegment(ctx context.Context, marker Marker, o
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/list-blobs.
 func (c ContainerURL) ListBlobsHierarchySegment(ctx context.Context, marker Marker, delimiter string, o ListBlobsSegmentOptions) (*ListBlobsHierarchySegmentResponse, error) {
 	if o.Details.Snapshots {
-		panic("snapshots are not supported in this listing operation")
+		sanityCheckFailed("snapshots are not supported in this listing operation")
```
devigned: This function already has an error on the return signature. Why not return an error here rather than a panic?

azblob/url_container.go
```diff
@@ -89,7 +86,7 @@ func (c ContainerURL) Create(ctx context.Context, metadata Metadata, publicAcces
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/delete-container.
 func (c ContainerURL) Delete(ctx context.Context, ac ContainerAccessConditions) (*ContainerDeleteResponse, error) {
 	if ac.IfMatch != ETagNone || ac.IfNoneMatch != ETagNone {
-		panic("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
+		sanityCheckFailed("the IfMatch and IfNoneMatch access conditions must have their default values because they are ignored by the service")
```
devigned: This function already has an error on the return signature. Why not return an error here rather than a panic?

azblob/zc_uuid.go
zezha-msft: Yes, this error practically never occurs. If we didn't want to panic on it, then we should simply ignore it, since the user wouldn't know how to deal with it even if we propagate it upwards.

azblob/zc_sas_account.go
zezha-msft: I'm a bit confused here. `perms.Parse` does not panic, and we already removed the panic here to return an error instead.

devigned: Please disregard. I must not have had my coffee yet.

azblob/url_container.go
zezha-msft: Great question! We perform a sanity check here because these parameters are not supported on the service end, and there could be potential data loss if the user relied on these parameters (and we let them).

Even if we returned this as an error instead, how could the user reasonably react to it? Would they call the API again without the offending access condition?

In my understanding, errors should be actionable. In this case, it is certainly not; so it might be better to panic and let the user catch this programmer error in their testing environment.

azblob/zc_sas_account.go
zezha-msft: No worries 😉

azblob/url_container.go
zezha-msft: The discussion will continue [here](https://github.com/Azure/azure-storage-blob-go/pull/73#discussion_r222069003).

azblob/url_container.go
zezha-msft: The discussion will continue [here](https://github.com/Azure/azure-storage-blob-go/pull/73#discussion_r222069003).

zezha-msft: The discussion will continue [here](https://github.com/Azure/azure-storage-blob-go/pull/73#discussion_r222069003).

azblob/url_container.go
zezha-msft: The discussion will continue [here](https://github.com/Azure/azure-storage-blob-go/pull/73#discussion_r222069003).

azblob/url_container.go
devigned: I agree the error should be actionable just the same way that you would get an error from parsing something. Of course, you could not retry parsing that same input and expect a different result, but the error is actually actionable and most of all, it does not require special handling to ensure the application does not crash by the consumer of your API.

I find this usage of panic particularly subjective.

If you would like to indicate some level of severity with the error, then I suggest making a more informative error type. That way, you could provide some indication on the error type similar to how `net.Error` https://golang.org/pkg/net/#Error indicates some category of errors are temporary.

marstr: We keep going around and around on this. By the widely excepted convention documented in [the golang wiki](https://github.com/golang/go/wiki/PanicAndRecover#usage-in-a-package), panic/recover is just a way to shortcut control flow inside of your package and shouldn't be exposed through package boundaries. While calling it "sanityCheckFailed" it at least spares folks the number of hits they see when they `grep panic` on this library, but the spirit of things doesn't change at all. Simply put, the community does not recognize any subtlety to when to return an error and when to panic: there's just a dangerous one and a safe one. Whether or not we agree with this assessment is not really relevant. 

In all cases where the function already ready returns an error, at the moment is is absolutely the correct thing to use that error.

zezha-msft: Hi @devigned, I'm a bit confused by what you mean here:
> Of course, you could retry parsing the same input, but you will still get the same error. The error is actually actionable. Most of all, it does not require special handling to ensure the application does not crash by the consumer of your API.

To clarify, the receiver `SetMetadata` manipulates metadata on the container, but the Storage service ignores certain container access conditions(IfUnmodifiedSince, IfMatch, and IfNoneMatch). If the user relied on these access conditions, there could be data corruption and it's very hard to recover their data when they discover this problem. In other words, we are not doing some generic parsing here; we are trying to help the user not corrupt their data. Furthermore, the same problem exists with 2 other APIs, including `Delete` on `ContainerURL`, which ignores IfMatch and IfNoneMatch. We certainly do not want the user to mistakenly delete their entire container. 

If we take a step back, there are indeed other ways of helping the user with this problem. The two possibilities that I see are:

1. Return an error instead of panicking.
2. Make these 3 receivers on `ContainerURL` take in individual parameters instead of `ContainerAccessConditions`.

Would you prefer option 1 or 2? And why?

zezha-msft: Hi @marstr, thanks for the review. We indeed agreed to stop debating on whether panics are good/bad; instead we should be looking at our usage of panics on a case by case basis, and see if there's a way of removing it without impacting the quality/safety of the SDK. I certainly hope that you could see the progress we've made so far(there are only 13 calls to `sanityCheckFailed`). And with your help, I also hope to merge/publish this PR as soon as possible, to alleviate customer pain.

Here are the 13 calls to `sanityCheckFailed`.
image

1. 4 of these calls are on `ContainerURL`. They panic when the user relies on access conditions or parameters that are silently ignored by the Storage service. We are trying to protect the customers from data corruption/loss with the panics. We obviously do not expect the customers to write recover code for these APIs, we instead expect their code to fail immediately in the test environment when they attempt to use them.
2. 1 such call is on `SharedKeyCredential`'s `buildCanonicalizedResource`. When failing to parse the query parameters, we panic instead of throwing an error because there must be a serious problem with the SDK itself if we are unable to parse the query parameters that own our code added. In other words, we can no longer proceed if we cannot trust the protocol layer.
3. Another call is on `tokenCredential`'s constructor. We panic if the user attempts to use HTTP instead of HTTPS. This is because the user is at the risk of exposing their bearer token with HTTP. Instead of letting the call go out, we want the user to catch this in their test environment.
4. The next 2 calls are on the `unmap` API of the memory-mapped-file types(Unix and Windows). We've already talked about this usage, since if we cannot unmap a memory-mapped-file, we cannot proceed as there could be memory corruption.
5. The next call is in the retry policy. We panic when we are unable to rewind the body for a retry. This is a serious problem since the Body is a ReadSeeker; if we cannot rewind it for a retry, there could be data corruption if we continued. We can indeed return an error here instead of panicing, but what is the user supposed to do with this error? The last 3 panics on the list are also about seeking failures on the body.
6. The other call in the retry policy is when the response is nil while there is no error. This one is a consistency check, and basically should never happen. If we didn't check for it though, it will panic on the next line.

The above is a brief explanation for each remaining usage of panic. Each case has some nuances and took us a good amount of time to consider their validity. Please feel free to reach out if you want to discuss them.

In short, the statement that panics means unsafe code is simply not true. At least the usage number 4 proves it. If we cannot unmap a memory-mapped-file, why shouldn't we crash with a clear message? The other two options in this case are:
1. Remove the check for error
  This makes no sense, memory could be corrupted.
2. Return the error to the user 
  What are they supposed to do with it? Do you really expect them to write code to detect this?

Therefore, panics are in fact a necessity in some cases.

devigned: @zezha-msft, let me try to clarify. You previously stated:
> Even if we returned this as an error instead, how could the user reasonably react to it? Would they call the API again without the offending access condition?

I was giving an example of where an error is returned, which provides information to the user about what went wrong (actionable), but should not be retried. For example, when parsing / unmarshalling JSON into a struct the error returned is actionable, but you will receive the same result if you call unmarshal as second time. By no means should the code below panic (though it does under the covers, I believe, but is handled before it hits the package boundary).

```go
err=json.Unmarshal(content, &structure)
if err!=nil{
    fmt.Print("Error:",err)
}
```

In the case of `SetMetadata` and similar methods, if you return an error to the consumer, you'd have the same API protecting effect as the panic, but without the undesirable possibility of crashing the consumers application. Since you'd return the error after you check if etags exist, you have ensured the user will not be able to call the API (`c.client.SetAccessPolicy`) and possibly cause data loss.

So the short answer is that I would select __option 1__. Return the error. Provide adequate information in the error for the consumer to understand the reason for the error and how they would change their code to avoid the error in the future.

zezha-msft: Hi @devigned, thanks for the explanation! I'm ok with removing the panics in this case. 

Could you please also take a look at the other cases that I've mentioned above? And see if you have any thoughts on them.

devigned: Thank you, @zezha-msft.

- `ContainerURL`
  - Return an error. I would use the same logic applied to `SetMetadata` and alike.
- `buildCanonicalizedResource` and `SharedKeyCredential`
  - Return an error. Since these are part of the generated code and thus should be verifiable through tests produced by the generator, I would not panic, but ensure the generated code or the generator has sufficient tests to ensure this error will not occur. Furthermore, the `pipeline.PolicyFunc(func(ctx context.Context, request pipeline.Request) (pipeline.Response, error)` allows for an error return, so returning an error from `f.buildStringToSign` will not cause much pain for upward error propagation.
- `tokenCredential`
  - Return an error. I would use the same logic applied to `SetMetadata` and alike.
- `unmap`
  - Panic. I'm of two minds on this one. I don't like the panic, but something is wrong at the OS level. I'm ok with bailing out on this one. @marstr, @bketelsen or @erikstmartin what do you think.
- `ReadSeeker`
  - Return an error explaining that the consumer must supply a ReadSeeker. I'd use the same logic applied to `SetMetadata`
- `response == nil`
  - Return an error explaining what happened and to file an issue in this repo. The function you are in already has an error return, `func(ctx context.Context, request pipeline.Request) (response pipeline.Response, err error)`. If you'd like to have a stack for these error, consider using or building something like https://godoc.org/github.com/pkg/errors#WithStack.

zezha-msft:
@devigned @marstr I've updated the PR according to your suggestions. Please take another look and sign off if appropriate. Thanks!

azblob/url_container.go
marstr: > - unmap
>   - Panic. I'm of two minds on this one. I don't like the panic, but something is wrong at the OS level. I'm ok with bailing out on this one. @marstr, @bketelsen or @erikstmartin what do you think.

I agree, this seems like an acceptable place to panic.

PR #75, opened 2018-10-13
Fix off by one error in UploadStreamToBlockBlob

ncw:
Before this fix, if you tried to upload a single part blob with
UploadStreamToBlockBlob of exactly BufferSize you would get this
error:

===== RESPONSE ERROR (ServiceCode=Md5Mismatch) =====
Description=The MD5 value specified in the request did not match with the MD5 value calculated by the server.
RequestId:5bca9423-f01e-003f-550d-632dc8000000
Time:2018-10-13T15:55:57.7806913Z, Details:
   Code: Md5Mismatch
   ServerCalculatedMd5: 1B2M2Y8AsgTpgAmY7PhCfg==
   UserSpecifiedMd5: DzQ7CTESaiDxM9Z8KwGKOw==

This is caused by the uploadStreamToBlockBlobOptions.chunk() calculating
that more than one block was needed in this case.

### DIFF
--- a/azblob/highlevel.go
+++ b/azblob/highlevel.go
@@ -407,7 +407,7 @@ func (t *uploadStreamToBlockBlobOptions) start(ctx context.Context) (interface{}
 }
 
 func (t *uploadStreamToBlockBlobOptions) chunk(ctx context.Context, num uint32, buffer []byte) error {
-	if num == 0 && len(buffer) < t.o.BufferSize {
+	if num == 0 && len(buffer) <= t.o.BufferSize {
 		// If whole payload fits in 1 block, don't stage it; End will upload it with 1 I/O operation
 		t.firstBlock = buffer
 		return nil


zezha-msft:
Hi @ncw, thank you for this PR!

If you don't mind, could you please rebase this change to target the dev branch instead? Thank you!

ncw:
I've rebased that on dev (which seems to have confused GitHub somewhat!).

ncw:
> Thanks @ncw! I changed the target branch and it's fine now.

I didn't know you could do that  :-)

Thanks for merging.

PS Might I suggest that you write in the README that you expect pull requests on the `dev` branch?

zezha-msft:
Sorry @ncw, I was too hasty in my review. There's actually no off by 1 error, because if the payload is the exact same size as the buffer, it is possible that more data is coming. I've documented it in #80.

ncw:
@zezha-msft without that fix, if I send an input of exactly BufferSize to UploadStreamToBlockBlob it fails with the error I posted in the original posting: `The MD5 value specified in the request did not match with the MD5 value calculated by the server.`

Can you suggest a better fix?

zezha-msft:
Hi @ncw, yes you are right. I've submitted a separate PR(#81) for that fix. Please take a look if you have time. 😄

PR #76, opened 2018-10-13
build: fix canonical path for golint

ncw:
This stops the build erroring with

package github.com/golang/lint/golint: code in directory /go/src/github.com/golang/lint/golint expects import "golang.org/x/lint/golint


zezha-msft:
@ncw thank you!

PR #87, opened 2018-11-09
Allowed retryReader to report the failures that it detects

JohnRusk:
Adds an optional notification func (callback) that can be supplied in the RetryReaderOptions.  If supplied, the notification function is called for both retryable and non-retryable failures.  

The intent is that code consuming this package will supply the callback if it wants to log information about failures that are retried by RetryReader.


zezha-msft:
Included in #90

PR #90, opened 2018-12-07
Feature/retry reader supports forced retries

JohnRusk:
Allows programmatically forced retries (to deal with abnormally slow body reads).


JohnRusk:
Note to reviewers: this change to retry reader (for body reads) relies on zc_retry_policy (for retries of GET/PUT, not including body). Specifically, it relies on the fact that zc_retry_policy sets up a _per-request_ context, that is a child of the overall context. That's important, because cancelling the _overall_ context would be bad - it could terminate more than just the one request that we want to retry.  We only want to cancel the per-request context.

The dependency on zc_retry_policy is made somewhat explicit by looking for the new RequestCanceller interface on the Response body.  That points into the per-request context.  If that interface is not implemented on the Response body, then retryReader cannot support cancellation and will return nil from getForceRetryFuncOrNil.  It's up to the caller to check for the nil value. (Should we return a separate error instead?)

If we didn't want to rely on zc-retry-policy, we would have to find another way to create per-request contexts - both for retryReader's initialResponse and for any new requests created by retryReader's getter func.  If you feel we should look into that approach, to decouple us from zc-retry-policy, pls include that in your PR feedback. (Suggestions on _how_ to do so are welcome!)

zezha-msft:
The dev branch was out of sync with the master for some reason. I'll push review&push your changes to a branch on this repo to trigger the CI.

azblob/zc_policy_retry.go
```diff
@@ -296,6 +296,12 @@ func (rc *contextCancelReadCloser) Close() error {
 	return err
 }
 
+func (rc *contextCancelReadCloser) CancelRequest() {
```
zezha-msft: Instead of adding a new receiver, we could potentially just call `Close()`. It cancels the context in addition to closing the connection, which is desired.

zezha-msft: If we closed the connection while another go-routine is reading from the body, I'd expect the `Read` function to return an error and say that the connection broke, but certain bytes were read. This should be double-checked.

azblob/zc_retry_reader.go
```diff
@@ -120,3 +140,57 @@ func (s *retryReader) Close() error {
 	}
 	return nil
 }
+
+// Returns a function that can be used to force a retry within (and transparently to) an call to Read
+func (s *retryReader) getForceRetryFuncOrNil() func(){
```
zezha-msft: This is very specific to the use case in AzCopy, I'm debating whether this should be kept out of the SDK.

azblob/zc_retry_reader.go
```diff
@@ -53,14 +54,32 @@ type retryReader struct {
 	countWasBounded bool
 	o               RetryReaderOptions
 	getter          HTTPGetter
+
+	// forced re-read handling (which is the only thread-safe part of this type)
+	rrLock      *sync.Mutex
```
zezha-msft: Do we absolutely need a lock? 

I assume the main purpose is to protect `forceRetry()`, which must set `rrForced` and also cancel the HTTP Body at the same time. Then `wasForcedRetry()` could return the right value. But the lock doesn't seem to solve the race condition completely, because what if an error that shouldn't be retried is returned from `s.response.Body.Read(p)`,  and then `rrForced` is set to true; by the time we get to line 121 and call `wasForcedRetry()`, we'd end up retrying an error that shouldn't be retried.

zezha-msft:
Sorry for messing up the base branch. I've done the rebasing work and pushed to this [new branch](https://github.com/Azure/azure-storage-blob-go/tree/retry-reader-notifier). We can put the comments here and make the changes there.

azblob/zc_policy_retry.go
JohnRusk: I tried that, expecting the same behavior that you mentioned. And the first time it worked great... But, the next time I ran it I got about 6 "context canceled" errors.  I suspected some kind of race condition, but did not identify exactly where it might have been.  That might be debuggable and fixable but..

.. then I realized something else.  If we use `Close()`, we're actually making an assumption there. We are assuming that (a) closing this particular response Body will actually cancel the context and (b) the context that gets cancelled is one that was created uniquely for this request (i.e. it's a child context, not a parent one).  Those assumptions are true for bodies that are implemented by `contextCancelReadCloser` but not necessarily true for all hypothetical response bodies.  So... using the explicit RequestCanceller interface provides a check that our assumptions are met.  Why is that check useful?  Because someone may one day use retryReader without policyRetry, and the assumptions would then be no longer met.  With the explicit reliance on `contextCancelReadCloser` such a code change would be immediately detected as an error - because suddenly retryReader would be using a request body that did not implement the interface.  We can't get the same effect with `Close()` because _all_ bodies are closeable.

azblob/zc_retry_reader.go
JohnRusk: In hindsight, I think the signature should have been 
```func (s *retryReader) getForceRetryFunc() (func(), error) { ```

That's more idomatic in terms of the error handling.  Do you think that would make it more suitable for inclusion in the SDK?  If not, what would be a good way to take it out of the SDK?

Oh, BTW, the reason I made the method private was to avoid cluttering up the SDK too much. Being private, its only exposed indirectly in the SDKs public API, in the ```BodyWithForceableRetry``` method in response_helpers.

File: azblob/zc_retry_reader.go (hunk 3 above)
Comment (JohnRusk) [reply azblob/zc_retry_reader.go]: Yes, the use in `forceRetry()` is its main purpose.  And the lock avoids the scenario where `wasForcedRetry()` returns false even "after" the HTTP body has been cancelled.  So the lock correctly protects us from thinking we were _not_ cancelled when really we _were_.  That protection is important because, without it, we could occasionally misinterpret forced retries as a non-retryable error, and fail the transfer.  I.e. we fail a transfer that should _not_ have failed.

As you say, the lock doesn't protect us from the opposite situation, i.e.:  thinking we _were_  cancelled when really we _weren't_.  But IMHO the consequences of that are less severe than failing the transfer. The consequence is just that we retry the GET/read sequence.  Presumably that retry would fail for the same fatal reason, and then we'd (correctly) identify it as non-retryable. I.e. we fail a transfer that _should_ have failed. We just take a little longer to do it.

To solve the latter situation the only option I could think of was to check err in `wasForcedRetry` to see if it looks like a cancellation error.  But even that is not technically perfect, and the context cancelled error looks like something that we'd have to check for by its string message/content, which seemed a bit ugly.  Any suggestions a on better way to solve the latter issue, and how badly we need to do so?

File: azblob/zc_retry_reader.go [hunk 4]
```diff
@@ -36,6 +36,9 @@ type RetryReaderOptions struct {
 	MaxRetryRequests   int
 	doInjectError      bool
 	doInjectErrorRound int
+
+	// Is called, if non-nil, after any failure to read. Expected usage is diagnostic logging.
+	NotifyFailedRead  func(failureCount int, lastError error, willRetry bool)
```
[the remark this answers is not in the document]
JohnRusk: Done

azblob/zc_retry_reader.go
JohnRusk: We agreed, in verbal discussion, that is good enough as is.

azblob/zc_policy_retry.go
JohnRusk: @zezha-msft I've made and pushed the changes to simplify it.  As discussed offline, and documented in the new commit, early Closing does appear to be safe. It works independently of the context cancel mechanism, but looking through the Go SDK source code indicates that it has the same effect.

azblob/zc_retry_reader.go
JohnRusk: With recent refactoring, this routine is no longer required

zezha-msft:
Duplicate of #95.

PR #95, opened 2019-01-04
Feature/retry reader supports forced retries

zezha-msft:


zezha-msft:
It's looking good to me. I'll let @jiacfan take a look too before I merge.

azblob/zc_policy_request_log.go
```diff
@@ -110,7 +110,7 @@ func NewRequestLogPolicyFactory(o RequestLogOptions) pipeline.Factory {
 }
 
 // redactSigQueryParam redacts the 'sig' query parameter in URL's raw query to protect secret.
```
jiacfan: It would be good to update the comments for doc purpose if wish to expose it as public method.

azblob/zc_retry_reader.go
```diff
@@ -45,19 +64,37 @@ type RetryReaderOptions struct {
...
+		responseMu:  &sync.Mutex{},
+		o:           o}
 }
 
+
+func (s *retryReader) setResponse(r *http.Response){
+	s.responseMu.Lock()
+	defer s.responseMu.Unlock()
+	s.response = r
+}
```
jiacfan: Mino: Please ensure format is good, e.g. additional blank lines are avoided.

azblob/zc_retry_reader.go
```diff
@@ -45,19 +64,37 @@ type RetryReaderOptions struct {
...
+		countWasBounded: info.Count != CountToEnd,
+		response:    initialResponse,
+		responseMu:  &sync.Mutex{},
+		o:           o}
 }
 
+
+func (s *retryReader) setResponse(r *http.Response){
+	s.responseMu.Lock()
+	defer s.responseMu.Unlock()
```
jiacfan: Seems defer is not necessary here. What about use following to save per cost brought by defer?

s.responseMu.Lock()
s.response = r
s.responseMu.Unlock()

JohnRusk: Could do it that way.  Defer seems nice and clean to read and perf cost is negligible compared to what the code is actually doing here (i.e. its just done a new HTTP request).  The cost of a defer will be many, many orders of magnitude smaller, so I'm inclined to leave this as is and consider this resolved.  Does that sound OK?

azblob/zc_policy_request_log.go
JohnRusk: Done

azblob/zc_retry_reader.go
JohnRusk: Done. (Ran gofmt, which I thought had run before, but clearly it hadn't).

JohnRusk:
Yes @jiacfan I'll help with getting them into the File SDK tomorrow. Just waiting on @zezha-msft to merge this one first.

zezha-msft:
Done! Thanks @JohnRusk!

PR #109, opened 2019-03-25
Issue 89: Making HTTPSender configurable in azblob pipeline

sumit-kalra:
Adding HTTPSender field to azblob's PipelineOptions in order to enable callers to specify a custom HTTPSender that can override the default used by the pipeline.

### DIFF
--- a/azblob/zc_pipeline.go
+++ b/azblob/zc_pipeline.go
@@ -17,6 +17,9 @@ type PipelineOptions struct {
 
 	// Telemetry configures the built-in telemetry policy behavior.
 	Telemetry TelemetryOptions
+
+	// HTTPSender configures the sender of HTTP requests
+	HTTPSender pipeline.Factory
 }
 
 // NewPipeline creates a Pipeline using the specified credentials and options.
@@ -38,5 +41,5 @@ func NewPipeline(c Credential, o PipelineOptions) pipeline.Pipeline {
 		pipeline.MethodFactoryMarker(), // indicates at what stage in the pipeline the method factory is invoked
 		NewRequestLogPolicyFactory(o.RequestLog))
 
-	return pipeline.NewPipeline(f, pipeline.Options{HTTPSender: nil, Log: o.Log})
+	return pipeline.NewPipeline(f, pipeline.Options{HTTPSender: o.HTTPSender, Log: o.Log})
 }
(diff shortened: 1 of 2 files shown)


zezha-msft:
Thanks @sumit-kalra! I'll review your change shortly.

azblob/zt_examples_test.go
```diff
@@ -154,6 +154,29 @@ func ExampleNewPipeline() {
...
+				// You can optionally use po (PolicyOptions) in this func.
+
+				// Forward the request to the next node in the pipeline:
+				response, err = next.Do(ctx, request)
+
+				// Process the response here. You can deserialize the body into an object.
+				// If you do this, also define a struct that wraps an http.Response & your
+				// deserialized struct. Have your wrapper struct implement the
+				// pipeline.Response interface and then return your struct (via the interface)
+				// After the pipeline completes, take response and perform a type assertion
```
zezha-msft: I'm a bit confused by this part. Could you please rephrase it?

zezha-msft:
Thanks @sumit-kalra for your contribution, I'll let @JeffreyRichter take a look too before I merge it in.

azblob/zt_examples_test.go
JeffreyRichter: Yes, this is just the sample code from the pipeline package. It is not an example of an HTTP Sender. Does this run? I'd expect it to crash when it calls next.Do since next should be null. The HTTP sender policy must not call next.Do; it should call Do on an HTTP client instead.

azblob/zc_pipeline.go
```diff
@@ -38,5 +41,5 @@ func NewPipeline(c Credential, o PipelineOptions) pipeline.Pipeline {
 		pipeline.MethodFactoryMarker(), // indicates at what stage in the pipeline the method factory is invoked
 		NewRequestLogPolicyFactory(o.RequestLog))
 
-	return pipeline.NewPipeline(f, pipeline.Options{HTTPSender: nil, Log: o.Log})
+	return pipeline.NewPipeline(f, pipeline.Options{HTTPSender: o.HTTPSender, Log: o.Log})
```
JeffreyRichter: This looks good

azblob/zt_examples_test.go
sumit-kalra: No, this does not run and , yes, this is verbatim sample from the pipeline package. Let me simplify this.

sumit-kalra: Okay, this has now been cleaned up.

zezha-msft:
Thanks @sumit-kalra for the PR. And thanks @JeffreyRichter for the review!

PR #111, opened 2019-03-28
Docs: update stdout for upload example

ehotinger:
For small chunks copy/pasting this example will result in `Wrote x of 0 bytes.`

`Len()` returns the number of bytes of the unread portion of the string, `Size()` returns the length of the underlying string.


zezha-msft:
@ehotinger thank you!

Issue #96, opened 2019-01-21
install error on Solaris


lhoward:
Similar issue to #23 on Solaris. Worked around by creating `azblob/zc_mmf_solaris.go` with following contents (from some searching/copy/pasting – I don't know Go).

```go
// +build solaris

package azblob

import (
        "os"
        "syscall"
        "golang.org/x/sys/unix"
)

type mmf []byte

func newMMF(file *os.File, writable bool, offset int64, length int) (mmf, error) {
        prot, flags := unix.PROT_READ, syscall.MAP_SHARED // Assume read-only
        if writable {
                prot, flags = syscall.PROT_READ|syscall.PROT_WRITE, syscall.MAP_SHARED
        }
        addr, err := unix.Mmap(int(file.Fd()), offset, length, prot, flags)
        return mmf(addr), err
}

func (m *mmf) unmap() {
        err := unix.Munmap(*m)
        *m = nil
        if err != nil {
                panic("if we are unable to unmap the memory-mapped file, there is serious concern for memory corruption")
        }
}
```

zezha-msft:
Hi @lhoward, thanks for reaching out!

I've logged this item to provide support for solaris.

lhoward:
Re-reading that patch, and with about 1% more Go knowledge than when I filed the report, it should not mix the syscall and unix packages (so, remove import syscall and s/syscall/unix/).

petems:
I think we can just add the solaris build tag to the existing file right?

Opened a PR to do so 👍

lhoward:
No, it's a new file that uses `golang.org/x/sys/unix` instead of the `syscall` package; at least on the version of Go I have, syscall.Mmap and syscall.Munmap are unavailable on Solaris. The version below avoids mixing the two packages.

```go
// +build solaris

package azblob

import (
        "os"
        "golang.org/x/sys/unix"
)

type mmf []byte

func newMMF(file *os.File, writable bool, offset int64, length int) (mmf, error) {
        prot, flags := unix.PROT_READ, unix.MAP_SHARED // Assume read-only
        if writable {
                prot, flags = unix.PROT_READ|unix.PROT_WRITE, unix.MAP_SHARED
        }
        addr, err := unix.Mmap(int(file.Fd()), offset, length, prot, flags)
        return mmf(addr), err
}

func (m *mmf) unmap() {
        err := unix.Munmap(*m)
        *m = nil
        if err != nil {
[...]

PR #116, opened 2019-04-02
Add Solaris to build tags

petems:
Fixes https://github.com/Azure/azure-storage-blob-go/issues/96

### DIFF
--- a/azblob/zc_mmf_unix.go
+++ b/azblob/zc_mmf_unix.go
@@ -1,4 +1,4 @@
-// +build linux darwin freebsd openbsd netbsd
+// +build linux darwin freebsd openbsd netbsd solaris
 
 package azblob
 


zezha-msft:
Hi @petems, thanks for the PR! But it doesn't seem to work. If I run ` GOOS=solaris go build ./azblob`, I get:

```
azblob/zc_mmf_unix.go:17:15: undefined: syscall.Mmap
azblob/zc_mmf_unix.go:22:9: undefined: syscall.Munmap
```
I haven't looked too closely, but the syscall Mmap seems to be [missing](https://golang.org/src/syscall/syscall_solaris.go).

petems:
Damn, more complicated than I thought. Doing a quick google shows that you can switch fully to the `golang.org/x/sys/unix` libary over syscall and it'll work on Solaris

Let me open a new PR to do that, and add Solaris to the build targets

PR #117, opened 2019-04-03
Adds solaris support

petems:
* Switch from syscall to unix package which works cross platform

Closes https://github.com/Azure/azure-storage-blob-go/issues/96

Surplants #116


lhoward:
I am no Go expert, but I imagine you want to continue using the native syscall package on platforms other than Solaris.

lhoward:
Perhaps best to check in with someone that knows better than I! Otherwise, see the patch in the original issue #96 which I have revised.

petems:
I'm no go expert but looking at other repos they seem to have switched whole hog over to the new libray with no issues on all platforms, not just solaris (eg. https://github.com/prometheus/tsdb/pull/200/files)

But I'm happy with whatever choice is given 😄

zezha-msft:
Thank you @petems for following up on this PR! And thanks @lhoward for providing insights!

I'll try to enlist our Go expert's help, since I don't really have the bandwidth right now to investigate the implications of switching to `golang.org/x/sys/unix`. @JeffreyRichter could you please take a look?

lhoward:
Switching for Solaris only is the most risk averse, given it didn't work on Solaris in the first place. But, I leave it to the Go experts.

zezha-msft:
@lhoward I tend to agree with you that it's definitely less risky to only use `golang.org/x/sys` for solaris, but the syscall package is [deprecated](https://golang.org/pkg/syscall/). So we should be moving to `golang.org/x/sys` down the line.

JeffreyRichter:
I think it’s fine to use https://godoc.org/golang.org/x/sys/unix
And https://godoc.org/golang.org/x/sys/windows
(both of which use syscall under the convers).

My only concern is that these packages are very large and so the Go-compiled binary might be much larger than it is now. So, it is worth a test to see how badly this affect the final file.

If the final file is large, there are things that can be done to shrink it like using the -s & -w flags as described here<https://golang.org/cmd/link/>. This is probably worth doing anyway.


  *   Jeffrey Richter (watch Architecting Distributed Cloud Apps<https://azure.microsoft.com/training/distributed-cloud-apps/>)

zezha-msft:
Thanks! @JeffreyRichter

JohnRusk:
Ooops, I see this has dropped of our radar.  We'll get back to it, but won't be for a few weeks.

KodyKantor:
Hello. I got here from #96 because MinIO is failing to build on illumos due to this problem. Please do let me know if I can help at all with this effort.

JohnRusk:
@KodyKantor Do you mean MinIO is failing to build due to the error in this (the Azure Storage) SDK?

KodyKantor:
@JohnRusk yes, that's correct. It appears that MinIO pulls the azure-storage-blob-go/azblob package in and attempts to build it as part of its Azure Gateway mode of operation.

JohnRusk:
@JeffreyRichter You're looking after this SDK, right?  Pls see above.

JeffreyRichter:
I haven't done anything related to this since my initial research and response on this issue. I felt then that conditionally building with the https://godoc.org/golang.org/x/sys/unix, https://godoc.org/golang.org/x/sys/windows is fine.

Also, I think you should look into building AzCopy with the -s & -w flags as described here: https://golang.org/cmd/link/ to shrink binary size. Assuming you're not doing this already.

KodyKantor:
FWIW we had an illumos community member discover this issue and we've since been carrying the changes in `zc_mmf_unix.go`. The MinIO build and resulting binaries have been working just fine with that change in place.

I don't see a dramatic increase in binary size when moving from `syscall` to `...unix`, at least on my linux machine. Here's the azblob .a file first from master:
```
$ ls -l gopath/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
-rw-r--r-- 1 kkantor kkantor 6577886 Feb 27 09:10 gopath/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
```
And then using the `unix` import added in this patch:
```
$ ls -l gopath2/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
-rw-r--r-- 1 kkantor kkantor 6578422 Feb 27 09:12 gopath2/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
```

It looks like the .a file grew by 536 bytes, which is much less than a one percent increase.

The MinIO binary (a consumer of the azblob package) grew by 4KB moving from `syscall` to `unix`
```
$ ls -l gopath/bin
total 59948
-rwxr-xr-x 1 kkantor kkantor 61385916 Feb 27 09:29 minio
$ ls -l gopath2/bin/
total 59952
-rwxr-xr-x 1 kkantor kkantor 61390012 Feb 27 09:28 minio
```
In my opinion we shouldn't worry about this binary size increase.

KodyKantor:
FWIW we had an illumos community member discover this issue and we've since been carrying the changes in `zc_mmf_unix.go`. The MinIO build and resulting binaries have been working just fine with that change in place.

I don't see a dramatic increase in binary size when moving from `syscall` to `...unix`, at least on my linux machine. Here's the azblob .a file first from master:
```
$ ls -l gopath/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
-rw-r--r-- 1 kkantor kkantor 6577886 Feb 27 09:10 gopath/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
```
And then using the `unix` import added in this patch:
```
$ ls -l gopath2/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
-rw-r--r-- 1 kkantor kkantor 6578422 Feb 27 09:12 gopath2/pkg/linux_amd64/github.com/azure/azure-storage-blob-go/azblob.a
```

It looks like the .a file grew by 536 bytes, which is much less than a one percent increase.

The MinIO binary (a consumer of the azblob package) grew by 4KB moving from `syscall` to `unix`
```
$ ls -l gopath/bin
total 59948
-rwxr-xr-x 1 kkantor kkantor 61385916 Feb 27 09:29 minio
$ ls -l gopath2/bin/
total 59952
-rwxr-xr-x 1 kkantor kkantor 61390012 Feb 27 09:28 minio
```
In my opinion we shouldn't worry about this binary size increase.

JeffreyRichter:
I agree, this size differential is negligible. Thanks for sharing numbers

JohnRusk:
Thanks guys for talking this through.  Useful and informative.

Kody, the biggest thing you could do to help would be to review this PR. I'll invite you as a reviewer.  We generally try to get two reviewers for all PRs.  Once you've posted your review, please @mention me here and I'll try to arrange a second review.

@petems   I'm sorry this fell off our radar for so long.

Thanks @KodyKantor for bringing our attention back to it.

KodyKantor:
I pulled in this change locally and applied it on top of the latest master. Now we also need to update the 'azure-pipeline-go' dependency to v2.2 to get around mattn/go-ieproxy#14.

JohnRusk:
Kody, Just FYI, I'm unable to look at any of the above this week.  What do you need from those of us who have admin rights here (e.g. to approve PRs), and are you able to use local builds and branches since we can't get to this this week?

cc @adreed-msft

KodyKantor:
@JohnRusk, thanks for keeping an eye on this. I think there are a couple paths to move forward here.

If @petems would like to continue with this PR he (or an admin?) could apply [this patch](https://github.com/Azure/azure-storage-blob-go/files/4276357/0001-update-azure-pipeline-go-dep.txt) that I crafted and update this PR. Another approach would be that I can open a PR against this repo and continue driving. Otherwise we could integrate this as-is and I can send a follow-on patch to update the dependency.

> are you able to use local builds and branches since we can't get to this this week?

Yes, we've been carrying local patches for this that we can continue doing until we get this fixed.

petems:
@KodyKantor Applied patch! 🤝

petems:
I've also set the PR to allow edits from maintainers, so if anyone needs to change anything else, go for it 😄

KodyKantor:
Hi @JohnRusk and @JeffreyRichter, can we integrate this change soon? It looks good to me.

JohnRusk:
Hi @KodyKantor , I don't have any time this week, I'm sorry.  @zezha-msft any chance you can look at this?

PR #124, opened 2019-04-25
SAS snapshot support & test added

adreed-msft:
Well. The last try didn't come out super great, did it?

Let's go in for take 2.


azblob/parsing_urls.go
```diff
@@ -124,6 +124,10 @@ func (up BlobURLParts) URL() url.URL {
 
 	rawQuery := up.UnparsedParams
 
+	if up.Snapshot == "" && !up.SAS.snapshotTime.IsZero() {
```
zezha-msft: Please add a comment here to clarify the intent. 

If I understood it correctly, if the SAS targets a snapshot, and no snapshot ID was specified, we help the user by specifying it.

azblob/sas_service.go
```diff
@@ -30,7 +31,15 @@ type BlobSASSignatureValues struct {
 // the proper SAS query parameters.
 func (v BlobSASSignatureValues) NewSASQueryParameters(sharedKeyCredential *SharedKeyCredential) (SASQueryParameters, error) {
 	resource := "c"
-	if v.BlobName == "" {
+	if v.SnapshotTime.IsZero() {
```
zezha-msft: hmmm shouldn't it be `!v.SnapshotTime.IsZero()`?

azblob/zt_sas_blob_snapshot_test.go
```diff
@@ -0,0 +1,42 @@
+package azblob_test
+
+import(
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	chk "gopkg.in/check.v1"
+	"os"
+	"time"
+)
+
+func (s *aztestsSuite) TestNewSASQueryParametersSnapshot(c *chk.C) {
```
zezha-msft: Please exercise the SAS, make sure it's signed correctly. For example:

1. Create a blob + a snapshot
2. Generate a snapshot SAS and read the snapshot, make sure the snapshot SAS cannot ready the blob itself

zezha-msft:
The change seems to have broken the [CI](https://travis-ci.org/Azure/azure-storage-blob-go/builds/524660707?utm_source=github_status&utm_medium=notification)

azblob/zt_sas_blob_snapshot_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"bytes"
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	chk "gopkg.in/check.v1"
+	"strings"
+	"time"
+)
+
+func (s *aztestsSuite) TestSnapshotSASUsage(c *chk.C) {
+	//Generate URLs ----------------------------------------------------------------------------------------------------
+	bsu := getBSU()
```
zezha-msft: Great test!

Minor: I'd suggest to make the test more concise by using Shared Key for the set up and taking out the part where Account SAS is used, and make that a test for the account SAS instead.

azblob/parsing_urls.go
```diff
@@ -124,6 +124,11 @@ func (up BlobURLParts) URL() url.URL {
 
 	rawQuery := up.UnparsedParams
 
+	//If no snapshot is initially provided, fill it in from the SAS query properties to help the user
+	if up.Snapshot == "" && !up.SAS.snapshotTime.IsZero() {
+		up.Snapshot = up.SAS.snapshotTime.Format(SnapshotTimeFormat)
```
zezha-msft: FYI @rickle-msft @jaschrep-msft I remember we made the decision not to append the snapshot ID when the customer is asking for a SAS token. But in the Go SDK, the customer never asks for just the token, instead they ask for the whole URL. So here we help them by including the snapshot ID if they forgot to set it on the BlobURLParts.

Please let us know if you have any concern.

zezha-msft:
@rickle-msft could you please take a look?

azblob/parsing_urls.go
rickle-msft: I think this is fine. If they have a SAS that only goes to a snapshot, it seems reasonable that would help them towards targeting the resource they can actually access. I don't imagine anyone will complain that we didn't append their snapshot sas to a base blob and let the request fail.

azblob/zt_sas_blob_snapshot_test.go
```diff
@@ -0,0 +1,89 @@
...
+	downloadedData.ReadFrom(reader)
+	reader.Close()
+
+	c.Assert(data, chk.Equals, downloadedData.String())
+
+	//Try (and fail to) delete from snapshot ---------------------------------------------------------------------------
+	deleteResponse, err := sburl.Delete(ctx, azblob.DeleteSnapshotsOptionOnly, azblob.BlobAccessConditions{})
+	if err == nil { //This absolutely SHOULD error out.
+		c.Fatal(deleteResponse)
+	}
```
rickle-msft: I think it's a good idea to add the snapshot sas to the base blob url and ensure that operation fails. Otherwise I think it's possible that we have simply created a regular blob sas

azblob/zt_sas_blob_snapshot_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	downloadedData := &bytes.Buffer{}
+	reader := downloadResponse.Body(azblob.RetryReaderOptions{})
+	downloadedData.ReadFrom(reader)
+	reader.Close()
+
+	c.Assert(data, chk.Equals, downloadedData.String())
+
+	//Try (and fail to) delete from snapshot ---------------------------------------------------------------------------
+	deleteResponse, err := sburl.Delete(ctx, azblob.DeleteSnapshotsOptionOnly, azblob.BlobAccessConditions{})
```
rickle-msft: Discussed offline. This should actually succeed if we get rid of the snapshot options.

azblob/zt_sas_blob_snapshot_test.go
```diff
@@ -0,0 +1,107 @@
...
+	downloadedData := &bytes.Buffer{}
+	reader := downloadResponse.Body(azblob.RetryReaderOptions{})
+	downloadedData.ReadFrom(reader)
+	reader.Close()
+
+	c.Assert(data, chk.Equals, downloadedData.String())
+
+	//Try to delete snapshot -------------------------------------------------------------------------------------------
+	_, err = sburl.Delete(ctx, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+	if err != nil { //This absolutely SHOULD error out.
```
rickle-msft: We concluded that this part should be modified, correct? It should either be deleted since we already test the validity of the sas on the download or it should not expect an error as it is valid to delete a snapshot with a sas.

adreed-msft:
@rickle-msft Could you take another look? It's 90% just the test that matters. No snapshot options are provided on the delete, and an attempted use as a normal SAS token fails, as intended.

PR #125, opened 2019-04-30
SAS identity addition

adreed-msft:
These commits add the ability to generate and make use of identity-based SAS tokens.

### DIFF
--- a/azblob/sas_service.go
+++ b/azblob/sas_service.go
@@ -25,90 +25,27 @@ type BlobSASSignatureValues struct {
 	ContentEncoding    string // rsce
 	ContentLanguage    string // rscl
 	ContentType        string // rsct
-	UserDelegationKey  UserDelegationKey
-}
-
-func (v BlobSASSignatureValues) NewIdentitySASQueryParameters(AccountName string) (SASQueryParameters, error) {
-	if v.UserDelegationKey.Value == "" {
-		return SASQueryParameters{}, errors.New("please provide a UserDelegationKey")
-	}
-
-	resource := "b"
-	//Make sure the permission characters are in the correct order
-	perms := &BlobSASPermissions{}
-	if err := perms.Parse(v.Permissions); err != nil {
-		return SASQueryParameters{}, err
-	}
-	v.Permissions = perms.String()
-
-	if v.Version == "" {
-		v.Version = SASVersion
-	}
-	startTime, expiryTime := FormatTimesForSASSigning(v.StartTime, v.ExpiryTime)
-
-	udk := v.UserDelegationKey
-	udkStart, udkExpiry := FormatTimesForSASSigning(udk.SignedStart, udk.SignedExpiry)
-	udkString := strings.Join([]string{
-		udk.SignedOid,
-		udk.SignedTid,
-		udkStart,
-		udkExpiry,
-		udk.SignedService,
-		udk.SignedVersion,
-	}, "\n")
-
-	// String to sign: http://msdn.microsoft.com/en-us/library/azure/dn140255.aspx
-	stringToSign := strings.Join([]string{
-		v.Permissions,
-		startTime,
-		expiryTime,
-		getCanonicalName(AccountName, v.ContainerName, v.BlobName),
-		udkString,
-		v.IPRange.String(),
-		string(v.Protocol),
-		v.Version,
-		resource,
-		"",                   // signed timestamp, @TODO add for snapshot sas feature
-		v.CacheControl,       // rscc
-		v.ContentDisposition, // rscd
-		v.ContentEncoding,    // rsce
-		v.ContentLanguage,    // rscl
-		v.ContentType},       // rsct
-		"\n")
-	signature := v.UserDelegationKey.ComputeHMACSHA256(stringToSign)
-
-	p := SASQueryParameters{
-		// Common SAS parameters
-		version:     v.Version,
-		protocol:    v.Protocol,
-		startTime:   v.StartTime,
-		expiryTime:  v.ExpiryTime,
-		permissions: v.Permissions,
-		ipRange:     v.IPRange,
-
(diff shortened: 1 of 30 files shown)


adreed-msft:
Requires that we move the build system to go 1.12 for testing.

JohnRusk:
Are the changes to the generated code created from a swagger file?  If so, could the swagger file be added to source control?  When looking at diffs in generated code, it's nice to be able to see how the source of the generation (i.e. the swagger file) has changed.  

That won't help for this change (because we don't currently have the "before" swagger in source control) but it will help for future changes.

Virepri:
The changes did majorly come from a swagger file, I'll include it when I get in tomorrow.

azblob/sas_service.go
```diff
@@ -51,13 +60,35 @@ func (v BlobSASSignatureValues) NewSASQueryParameters(sharedKeyCredential *Share
 	}
 	startTime, expiryTime := FormatTimesForSASSigning(v.StartTime, v.ExpiryTime)
 
+	 /*
+	 http://www.thecodelesscode.com/case/84
```
JohnRusk: I haven't understood this comment. I don't understand which variable is being used for two things.

adreed-msft: This was a "messy" holdover from a previous change I'd made-- Initially I had NewSASQueryParameters merged for Identity SAS.

I'll remove this comment and include the swagger in a moment.

zezha-msft:
> Requires that we move the build system to go 1.12 for testing.

Why btw?

azblob/sas_service.go
```diff
@@ -24,6 +25,83 @@ type BlobSASSignatureValues struct {
 	ContentEncoding    string // rsce
 	ContentLanguage    string // rscl
 	ContentType        string // rsct
+	UserDelegationKey  UserDelegationKey
+}
+
+func (v BlobSASSignatureValues) NewIdentitySASQueryParameters(AccountName string) (SASQueryParameters, error) {
```
zezha-msft: Shouldn't the UDK be taken in as a parameter? Similar to how the sharedKeyCredential is passed into NewSASQueryParameters. Basically, the credential is taken in to sign the SAS.

azblob/sas_service.go
```diff
@@ -24,6 +25,83 @@ type BlobSASSignatureValues struct {
...
+		udk.SignedOid,
+		udk.SignedTid,
+		udkStart,
+		udkExpiry,
+		udk.SignedService,
+		udk.SignedVersion,
+	}, "\n")
+
+	// String to sign: http://msdn.microsoft.com/en-us/library/azure/dn140255.aspx
+	stringToSign := strings.Join([]string{
```
zezha-msft: This method feels very redundant, please consider doing some refactoring so that there's less duplicated code between this method and NewSASQueryParameters.

zezha-msft: Example [here](https://github.com/Azure/azure-storage-python/blob/master/azure-storage-blob/azure/storage/blob/sharedaccesssignature.py#L232).

azblob/zc_sas_query_params.go
```diff
@@ -190,6 +195,18 @@ func newSASQueryParameters(values url.Values, deleteSASParametersFromValues bool
 			p.contentLanguage = val
 		case "rsct":
 			p.contentType = val
+		case "skoid":
+			p.userDelegationKey.SignedOid = val
```
zezha-msft: these are SAS query parameters, no need for the intermediate `userDelegationKey`

azblob/zt_sas_identity_test.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"bytes"
+	"fmt"
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	chk "gopkg.in/check.v1"
+	"strings"
+	"time"
+)
+
+func (s *aztestsSuite) TestIdentitySASUsage(c *chk.C) {
```
zezha-msft: please consider adding some comments + grouping lines together to improve readability.

azblob/zt_sas_identity_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	//Create pipeline
+	p = azblob.NewPipeline(azblob.NewAnonymousCredential(), azblob.PipelineOptions{})
+
+	//Create blob SAS Identity URL
+	bSASParts := azblob.NewBlobURLParts(blobURL.URL())
+	bSASParts.SAS = bSAS
+	bsurl := bSASParts.URL()
+	fmt.Println(bsurl.String())
```
zezha-msft: please avoid print statements in tests.

zezha-msft:
@adreed-msft feel free to update [this line](https://github.com/Azure/azure-storage-blob-go/blob/master/.travis.yml#L3) to enable Go 1.12. I see that Autorest requires 1.12 now.

Btw usually it's a good idea to have the build passing before CR. :shipit:

azblob/zt_sas_identity_test.go
adreed-msft: That was leftover debug material. Thanks for catching it.

adreed-msft:
> Btw usually it's a good idea to have the build passing before CR. :shipit: 

Some would say this is self-evident. I explained why the build was failing if you didn't notice. 👀

azblob/zc_sas_query_params.go
adreed-msft: Originally, I had implemented this with the parameters rather than the intermediate userDelegationKey but found that the latter felt like a cleaner approach. I'm happy to revert, however.

azblob/sas_service.go
adreed-msft: Again, pulling in the previous ways I'd implemented this, I disliked that I was using signedIdentifier for two things. Again, I will revert this.

PR #131, opened 2019-05-23
User Delegation SAS Tokens

adreed-msft:
This set of commits adds support for user delegation SAS tokens.

In order to pass the tests, this module needs the following environment variables:
- `TENANT_ID` (defaults to `common`)
- `CLIENT_SECRET` or `OAUTH_TOKEN` (`OAUTH_TOKEN` must be in Azure/go-autorest/adal's json format)
- `APPLICATION_ID`

As a result, the test will fail until we supply these. @amishra-dev @JohnRusk you both *likely* know how to work with the CI better than I do (and probably actually have access unlike me), so, could you help with supplying these to CI?

As such, the tests support either injecting an existing OAuth token, or obtaining one as a service principal. The registered application needs to be able to access the (non HNS) storage account as a `Storage Blob Data Contributor` and a `Owner` in order to actually pass the test.

### DIFF
--- a/azblob/blob.json
+++ b/azblob/blob.json
@@ -1,0 +1,8009 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "Azure Blob Storage",
+    "version": "2018-11-09",
+    "x-ms-code-generation-settings": {
+      "header": "MIT",
+      "strictSpecAdherence": false
+    }
+  },
+  "x-ms-parameterized-host": {
+    "hostTemplate": "{url}",
+    "useSchemePrefix": false,
+    "positionInOperation": "first",
+    "parameters": [
+      {
+        "$ref": "#/parameters/Url"
+      }
+    ]
+  },
+  "securityDefinitions": {
+    "blob_shared_key": {
+      "type": "apiKey",
+      "name": "Authorization",
+      "in": "header"
+    }
+  },
+  "schemes": [
+    "https"
+  ],
+  "consumes": [
+    "application/xml"
+  ],
+  "produces": [
+    "application/xml"
+  ],
+  "paths": {},
+  "x-ms-paths": {
+    "/?restype=service&comp=properties": {
+      "put": {
+        "tags": [
+          "service"
+        ],
+        "operationId": "Service_SetProperties",
+        "description": "Sets properties for a storage account's Blob service endpoint, including properties for Storage Analytics and CORS (Cross-Origin Resource Sharing) rules",
+        "parameters": [
+          {
+            "$ref": "#/parameters/StorageServiceProperties"
+          },
+          {
+            "$ref": "#/parameters/Timeout"
+          },
+          {
+            "$ref": "#/parameters/ApiVersionParameter"
+          },
+          {
+            "$ref": "#/parameters/ClientRequestId"
+          }
+        ],
+        "responses": {
+          "202": {
+            "description": "Success (Accepted)",
+            "headers": {
+              "x-ms-request-id": {
+                "x-ms-client-name": "RequestId",
+                "type": "string",
+                "description": "This header uniquely identifies the request that was made and can be used for troubleshooting the request."
+              },
+              "x-ms-version": {
+                "x-ms-client-name": "Version",
+                "type": "string",
+                "description": "Indicates the version of the Blob service used to execute the request. This header is returned for requests made against version 2009-09-19 and above."
+              }
+            }
+          },
+          "default": {
+            "description": "Failure",
+            "headers": {
+              "x-ms-error-code": {
+                "x-ms-client-name": "ErrorCode",
+                "type": "string"
+              }
+            },
+            "schema": {
+              "$ref": "#/definitions/StorageError"
+            }
+          }
+        }
+      },
+      "get": {
+        "tags": [
+          "service"
+        ],
+        "operationId": "Service_GetProperties",
+        "description": "gets the properties of a storage account's Blob service, including properties for Storage Analytics and CORS (Cross-Origin Resource Sharing) rules.",
+        "parameters": [
+          {
+            "$ref": "#/parameters/Timeout"
+          },
+          {
+            "$ref": "#/parameters/ApiVersionParameter"
+          },
+          {
+            "$ref": "#/parameters/ClientRequestId"
+          }
+        ],
+        "responses": {
+          "200": {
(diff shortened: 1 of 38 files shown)


azblob/sas_service.go
```diff
@@ -28,9 +29,23 @@ type BlobSASSignatureValues struct {
 
 // NewSASQueryParameters uses an account's shared key credential to sign this signature values to produce
```
zezha-msft: Update doc string

azblob/url_service.go
```diff
@@ -27,6 +27,16 @@ func NewServiceURL(primaryURL url.URL, p pipeline.Pipeline) ServiceURL {
 	return ServiceURL{client: client}
 }
 
+//GetUserDelegationKey obtains a UserDelegationKey object using the base ServiceURL object.
```
zezha-msft: Mention that OAuth is needed for this API to work.

azblob/zt_test.go
```diff
@@ -232,6 +234,66 @@ func getGenericCredential(accountType string) (*azblob.SharedKeyCredential, erro
 	return azblob.NewSharedKeyCredential(accountName, accountKey)
 }
 
+func getOAuthCredential(accountType string) (*azblob.TokenCredential, error) {
+	oauthTokenEnvVar := accountType + "OAUTH_TOKEN"
```
zezha-msft: some comments here to explain the setup would be nice.

azblob/zt_user_delegation_sas_test.go
```diff
@@ -0,0 +1,90 @@
+package azblob_test
+
+import (
+	"bytes"
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	chk "gopkg.in/check.v1"
+	"strings"
+	"time"
+)
+
```
zezha-msft: Please add test for container.

azblob/zt_user_delegation_sas_test.go
```diff
@@ -0,0 +1,90 @@
...
+		c.Fatal(err)
+	}
+	err = reader.Close()
+	if err != nil {
+		c.Fatal(err)
+	}
+	c.Assert(data, chk.Equals, downloadedData.String())
+
+	//Delete the item using the User Delegation SAS URL; must succeed
+	_, err = bSASURL.Delete(ctx, azblob.DeleteSnapshotsOptionInclude, azblob.BlobAccessConditions{})
```
zezha-msft: nice test.

azblob/blob.json
```diff
@@ -0,0 +1,8009 @@
+{
```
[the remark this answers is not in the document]
adreed-msft: Just making the note that we discussed that John told me to check in the spec.

adreed-msft:
@JohnRusk @zezha-msft could you guys hit this PR with another CR? Thanks

azblob/user_delegation_credential.go
```diff
@@ -0,0 +1,39 @@
+package azblob
+
+import (
```
rickle-msft: Should this be a zc_ file? Not sure if we can make a user delegation sas for queues and files. The rest docs don't seem to be updated

azblob/zc_credential_shared_key.go
```diff
@@ -39,6 +39,14 @@ func (f SharedKeyCredential) AccountName() string {
 	return f.accountName
 }
 
+func (f SharedKeyCredential) getAccountKey() []byte {
+	return f.accountKey
+}
+
+func (f SharedKeyCredential) getUDKParams() *UserDelegationKey {
```
rickle-msft: I wonder if it would be helpful to include a comment here to say that this method is present to satisfy the StorageCredential interface. Unless it's understood in Go that basically no-op methods like this are generally added to satisfy an interface.

azblob/zc_sas_query_params.go
```diff
@@ -63,6 +63,36 @@ type SASQueryParameters struct {
 	contentEncoding    string      `param:"rsce"`
 	contentLanguage    string      `param:"rscl"`
 	contentType        string      `param:"rsct"`
+	signedOid          string
```
rickle-msft: Looks like we're missing the `param: xyz` tags for all these new parameters

azblob/user_delegation_credential.go
adreed-msft: From my understanding, User Delegation SAS is blob-exclusive at the moment and Ze doesn't think it'll be coming to files/ADLSG2.

azblob/sas_service.go
```diff
@@ -51,13 +56,31 @@ func (v BlobSASSignatureValues) NewSASQueryParameters(sharedKeyCredential *Share
 	}
 	startTime, expiryTime := FormatTimesForSASSigning(v.StartTime, v.ExpiryTime)
 
+	signedIdentifier := v.Identifier
+
+	udk := credential.getUDKParams()
+
+	if udk != nil {
```
zezha-msft: much cleaner! 👍

azblob/user_delegation_credential.go
```diff
@@ -0,0 +1,40 @@
...
+	h.Write([]byte(message))
+	return base64.StdEncoding.EncodeToString(h.Sum(nil))
+}
+
+// Private method to return important parameters for NewSASQueryParameters
+func (f UserDelegationCredential) getUDKParams() *UserDelegationKey {
+	return &f.accountKey
+}
+
+//TODO: Should we have mimic credential policies?
```
zezha-msft: No since the UDK is only used for signing SAS and not signing requests directly.

azblob/zt_user_delegation_sas_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	c.Assert(data.String(), chk.Equals, "hello world!")
+	_, err = bblob.Delete(ctx, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+	if err != nil {
+		c.Fatal(err)
+	}
+}
+
+// Creates a blob, takes a snapshot, downloads from snapshot, and deletes from the snapshot w/ the token
+func (s *aztestsSuite) TestUserDelegationSAS(c *chk.C) {
```
zezha-msft: Minor: rename to TestUserDelegationSASBlob

adreed-msft:
Squashing and merging.

PR #136, opened 2019-06-19
General secondary host improvements

adreed-msft:
So far:

- Fixed a nil pointer deref when pinging the secondary host
- Made it actually ping the secondary host when failing over

### DIFF
--- a/azblob/zc_policy_retry.go
+++ b/azblob/zc_policy_retry.go
@@ -177,6 +177,9 @@ func NewRetryPolicyFactory(o RetryOptions) pipeline.Factory {
 
 				if !tryingPrimary {
 					requestCopy.Request.URL.Host = o.retryReadsFromSecondaryHost()
+					requestCopy.URL.Host = o.retryReadsFromSecondaryHost()
+					requestCopy.Host = o.retryReadsFromSecondaryHost()
+					requestCopy.Request.Host = o.retryReadsFromSecondaryHost()
 				}
 
 				// Set the server-side timeout query parameter "timeout=[seconds]"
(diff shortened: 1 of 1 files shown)


zezha-msft:
Btw was this the bug that the customer (not to be mentioned here) reported? :shipit:

adreed-msft:
No, this was *not* the bug, but it was one I encountered whilst trying to figure out the customer's bug. I've managed to repro the customer's bug and I'm working on fixing it.

azblob/zc_policy_retry.go
```diff
@@ -177,6 +177,9 @@ func NewRetryPolicyFactory(o RetryOptions) pipeline.Factory {
 
 				if !tryingPrimary {
 					requestCopy.Request.URL.Host = o.retryReadsFromSecondaryHost()
+					requestCopy.URL.Host = o.retryReadsFromSecondaryHost()
```
JohnRusk: If I understand correctly, by the rules of struct embedding `requestCopy.URL` is an alias for `requestCopy.Request.URL`  so I think there should be just two lines here, instead of four. Namely, there should be 
```requestCopy.Request.URL.Host = o.retryReadsFromSecondaryHost()```
and 
```requestCopy.Request.Host = o.retryReadsFromSecondaryHost()```

But not the other two lines. (Alternatively, and with the exact same meaning, ".Request." could be omitted from both the lines I suggest, to produce identical code.  I.e. out of the 4 lines currently here, there are two equivalent pairs).  Can we please get this down to one of the two-line versions?

adreed-msft: Ah! Good point. I was simply running through Goland's code suggestions-- probably a better idea to just look at the struct itself.

zezha-msft:
:shipit:

adreed-msft:
Squashing and merging due to double approve.

Issue #139, opened 2019-07-16
Segfault when UploadFileToBlockBlob return errors


ramielrowe:
### Which version of the SDK was used?

8a1deeeabe0a24f918d29630ede0da2a1c8f3b2f

### Which platform are you using? (ex: Windows, Linux, Debian)

Linux

### What problem was encountered?

Errors can be returned by [doBatchTransfer](https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/highlevel.go#L298) before all goroutines have finished. There are two compounding reasons for this. First, this [cancel](https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/highlevel.go#L341) doesn't wait for anything to finish before returning. Second, the [context that is cancelled](https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/highlevel.go#L303), isn't actually used by anything. This ends up racing this [Munmap call](https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/highlevel.go#L150), which will unmap the file out from under active goroutines during errors. This ends up causing segfaults.

### How can we reproduce the problem in the simplest way?

Sample golang code to reproduce: https://gist.github.com/ramielrowe/cf41891d38fa9bf636b7f45eb81a0cd0#file-mmap_segfault-go

### Have you found a mitigation/solution?

Here's an example pattern that will safely return errors and run munmap.

https://gist.github.com/ramielrowe/cf41891d38fa9bf636b7f45eb81a0cd0#file-mmap_safe-go

zezha-msft:
Hi @ramielrowe, thanks for reaching out and providing such details repro steps!

I've logged this bug to be investigated/addressed.

svrc-pivotal:
We've been running into these segfaults fairly regularly on 0.7.0.

feiyushi:
We may have hit the same error but in a different method `DownloadBlobToFile`. 
This segfault has caused our containers to restart and thus unnecessarily prolonged the operation execution time and sometimes caused operation failures.
Here's a **unordered** call stack extracted from production log, hope it helps:
```
goroutine 1045 [running]:
panic: runtime error: invalid memory address or nil pointer dereference
u0009/vendor/github.com/Azure/azure-storage-blob-go/azblob/zz_response_helpers.go:62 +0x2dd
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x55c9f5f5e98d]
[...]

PR #145, opened 2019-08-13
Fixed Segfault in DoBatchTransfer

zezha-msft:
Fixing the issue #139, which had extremely clear feedback on what was wrong: 

1. DoBatchTransfer did not pass `ctx` along properly.
2. More importantly, DoBatchTransfer did not wait for all routines to stop before quitting, resulting in the memory-mapped file being closed while some routines are still working with it.

### DIFF
--- a/azblob/highlevel.go
+++ b/azblob/highlevel.go
@@ -66,7 +66,7 @@ func UploadBufferToBlockBlob(ctx context.Context, b []byte,
 	if o.BlockSize == 0 {
 		// If bufferSize > (BlockBlobMaxStageBlockBytes * BlockBlobMaxBlocks), then error
 		if bufferSize > BlockBlobMaxStageBlockBytes*BlockBlobMaxBlocks {
-			return nil, errors.New("Buffer is too large to upload to a block blob")
+			return nil, errors.New("buffer is too large to upload to a block blob")
 		}
 		// If bufferSize <= BlockBlobMaxUploadBlobBytes, then Upload should be used with just 1 I/O request
 		if bufferSize <= BlockBlobMaxUploadBlobBytes {
@@ -76,7 +76,7 @@ func UploadBufferToBlockBlob(ctx context.Context, b []byte,
 			if o.BlockSize < BlobDefaultDownloadBlockSize { // If the block size is smaller than 4MB, round up to 4MB
 				o.BlockSize = BlobDefaultDownloadBlockSize
 			}
-			// StageBlock will be called with blockSize blocks and a parallelism of (BufferSize / BlockSize).
+			// StageBlock will be called with blockSize blocks and a Parallelism of (BufferSize / BlockSize).
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


azblob/highlevel.go
```diff
@@ -285,64 +285,69 @@ func DownloadBlobToFile(ctx context.Context, blobURL BlobURL, offset int64, coun
...
 
 		operationChannel <- func() error {
-			return o.operation(offset, curChunkSize)
+			return o.Operation(offset, curChunkSize, ctx)
 		}
 	}
 	close(operationChannel)
 
 	// Wait for the operations to complete.
+	var firstErr error = nil
```
adreed-msft: The setup below looks good to me, but we should consider using a sync.WaitGroup here, maybe?

azblob/highlevel.go
```diff
@@ -285,64 +285,69 @@ func DownloadBlobToFile(ctx context.Context, blobURL BlobURL, offset int64, coun
...
+	Operation     func(offset int64, chunkSize int64, ctx context.Context) error
+	OperationName string
 }
 
-// doBatchTransfer helps to execute operations in a batch manner.
-func doBatchTransfer(ctx context.Context, o batchTransferOptions) error {
+// DoBatchTransfer helps to execute operations in a batch manner.
+// Can be used by users to customize batch works (for other scenarios that the SDK does not provide)
+func DoBatchTransfer(ctx context.Context, o BatchTransferOptions) error {
+	if o.ChunkSize == 0 {
```
adreed-msft: My chunk size can be negative?

Two ways to fix this:
1. ChunkSize should be a uint64
2. This should check if it's less than or equal to zero

azblob/zt_highlevel_test.go
```diff
@@ -329,3 +332,104 @@ func (s *aztestsSuite) TestDownloadBufferWithNonZeroOffsetAndCount(c *chk.C) {
...
+			// simulate doing some work (HTTP call in real scenarios)
+			// later chunks later longer to finish
+			time.Sleep(time.Second * time.Duration(offset))
+			// simulate having gotten data and write it to the memory mapped file
+			mmf.write("input")
+
+			// with one of the chunks, pretend like an error occurred (like the network connection breaks)
+			if offset == 3 {
+				return expectedFirstError
+			} else if offset > 3 {
```
JohnRusk: nice

azblob/zt_url_container_test.go
```diff
@@ -221,8 +221,6 @@ func (s *aztestsSuite) TestContainerDeleteIfModifiedSinceTrue(c *chk.C) {
 	bsu := getBSU()
 	containerURL, _ := createNewContainer(c, bsu)
 
-	defer deleteContainer(c, containerURL)
```
JohnRusk: Why no delete now?

zezha-msft: The test is already deleting it, I'm really puzzled by why it was working previously.

azblob/highlevel.go
zezha-msft: Hmm, we read the responses back synchronously though, why is the sync.WaitGroup needed?

azblob/zt_highlevel_test.go
zezha-msft: I applied TDD LOL

azblob/highlevel.go
zezha-msft: Good point, we should take another look and make all the ints unsigned. I'll ship like this now to unblock our internal cx.

PR #152, opened 2019-11-11
Expose GetAccountInfo to the user

adreed-msft:
This is required for azcopy to handle SetTier errors more gracefully.

### DIFF
--- a/azblob/url_service.go
+++ b/azblob/url_service.go
@@ -38,6 +38,11 @@ func (s ServiceURL) GetUserDelegationCredential(ctx context.Context, info KeyInf
 	return NewUserDelegationCredential(strings.Split(s.client.url.Host, ".")[0], *udk), nil
 }
 
+func (s ServiceURL) GetAccountInfo(ctx context.Context) (*ServiceGetAccountInfoResponse, error) {
+	accountInfo, err := s.client.GetAccountInfo(ctx)
+	return accountInfo, err
+}
+
 // URL returns the URL endpoint used by the ServiceURL object.
 func (s ServiceURL) URL() url.URL {
 	return s.client.URL()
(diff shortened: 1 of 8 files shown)


azblob/url_service.go
```diff
@@ -38,6 +38,11 @@ func (s ServiceURL) GetUserDelegationCredential(ctx context.Context, info KeyInf
 	return NewUserDelegationCredential(strings.Split(s.client.url.Host, ".")[0], *udk), nil
 }
 
+func (s ServiceURL) GetAccountInfo(ctx context.Context) (*ServiceGetAccountInfoResponse, error) {
```
zezha-msft: >GetAccountInfo [](start = 20, length = 14)

this API works on container and blob too

azblob/zt_url_service_test.go
```diff
@@ -9,6 +9,14 @@ import (
 	chk "gopkg.in/check.v1" // go get gopkg.in/check.v1
 )
 
+func (s *aztestsSuite) TestGetAccountInfo(c *chk.C) {
+	sa := getBSU()
+
+	// Ensure the call succeeded. Don't test for specific account properties because we can't/don't want to set account properties.
```
zezha-msft: >// Ensure the call succeeded. Don't test for specific account properties because we can't/don't want to set account properties. [](start = 1, length = 127)

perhaps validate that it's at least not empty

zezha-msft:
https://docs.microsoft.com/en-us/rest/api/storageservices/get-account-information#request

azblob/url_service.go
zezha-msft: https://docs.microsoft.com/en-us/rest/api/storageservices/get-account-information#request

go.mod
```diff
@@ -5,3 +5,5 @@ require (
 	github.com/kr/pretty v0.1.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
 )
+
```
JohnRusk: Minor: given that this is a public SDK, it would be better if any Go version change was in a separate PR, instead of here.

adreed-msft:
Squashing and merging.

PR #153, opened 2019-11-21
Added support for CopyBlobFromURL (sync) and upgrade version to 2019-02-02

zezha-msft:
Sorry for the large PR. It was necessary to generate the code as there were missing parameters from the Swagger.


azblob/url_block_blob.go
```diff
@@ -108,6 +115,23 @@ func (bb BlockBlobURL) GetBlockList(ctx context.Context, listType BlockListType,
...
+	dstIfModifiedSince, dstIfUnmodifiedSince, dstIfMatchETag, dstIfNoneMatchETag := dstac.ModifiedAccessConditions.pointers()
+	dstLeaseID := dstac.LeaseAccessConditions.pointers()
+
+	return bb.blobClient.CopyFromURL(ctx, source.String(), nil, metadata, AccessTierNone,
+		srcIfModifiedSince, srcIfUnmodifiedSince,
+		srcIfMatchETag, srcIfNoneMatchETag,
+		dstIfModifiedSince, dstIfUnmodifiedSince,
+		dstIfMatchETag, dstIfNoneMatchETag,
+		dstLeaseID, nil, srcContentMD5)
+}
```
adreed-msft: On the one hand, this seems questionable to have on a block blob URL, since according to the docs, it's under the general blob operations.

On the other hand, it can only target block blobs.

So this is mostly just a question of "how well can the user locate this?".

@JohnRusk thoughts?

azblob/zt_url_block_blob_test.go
```diff
@@ -144,6 +144,77 @@ func (s *aztestsSuite) TestStageBlockFromURL(c *chk.C) {
...
+	_, badMD5 := getRandomDataAndReader(16)
+	_, err = destBlob.CopyFromURL(ctx, srcBlobURLWithSAS, azblob.Metadata{}, azblob.ModifiedAccessConditions{}, azblob.BlobAccessConditions{}, badMD5)
+	c.Assert(err, chk.NotNil)
+
+	// Edge case 2: Not providing any source MD5 should see the CRC getting returned instead
+	resp, err = destBlob.CopyFromURL(ctx, srcBlobURLWithSAS, azblob.Metadata{}, azblob.ModifiedAccessConditions{}, azblob.BlobAccessConditions{}, nil)
+	c.Assert(err, chk.IsNil)
+	c.Assert(resp.Response().StatusCode, chk.Equals, 202)
+	c.Assert(resp.XMsContentCrc64(), chk.Not(chk.Equals), "")
+}
```
adreed-msft: I would ensure we have decent logging for when a user tries to copy a blob that's too large, to ensure we're not just handing back a raw 400 with no real info.

azblob/url_block_blob.go
zezha-msft: The location of the REST doc is probably wrong. It should be under Block Blob operations instead.

azblob/zt_url_block_blob_test.go
zezha-msft: hmm that feel too much like testing the service though...

zezha-msft:
@JohnRusk I'm merging this so that I can get started with some work on the AzCopy side, please let me know if you have any comments.

PR #169, opened 2020-04-08
Fix Go SDK handling of SAS timestamps

### LINKED ISSUE #168: SDK rewrites SAS timestamps if they aren't in an exact format, which breaks signature verification
0.7.0 (via AzCopy/Storage Explorer), also 0.8.0
Windows, though I'd expect this issue to be platform-independent
The Go SDK's SAS handling code rewrites the SAS timestamps, which causes signature verification on the server side to fail.
The List Service SAS endpoint (https://docs.microsoft.com/en-us/rest/api/storagerp/storageaccounts/listservicesas), as well as probably others, currently generate timestamps with sub-second precision.  Even if you pass an expiration date without sub-second precision to the API, the service generates the signature with a timestamp that uses sub-second precision.  This SAS is then unusable with the Go SDK (and by proxy, AzCopy and Azure Storage Explorer)
No mitigation, but I will be opening a PR to resolve this shortly.  The original timestamp string needs to be retained, instead of expecting a `Time` to roundtrip cleanly.

I found my way here via azure/azure-storage-azcopy#122!

AtOMiCNebula:
The Go SDK currently rewrites SAS timestamps to always match a specific format (`"2006-01-02T15:04:05Z"`).  This is a problem, because if the timestamp doesn't roundtrip cleanly (i.e. a timestamp not matching that exact pattern) will be unusable with the Go SDK, because the SAS will get rewritten without regenerating the signature.

This change updates `SASQueryParameters` and friends to retain timestamps as their original string values, instead of as a `Time`.  This should eliminate roundtripping issues, since the timestamps will no longer be rewritten.

I will note that given my API surface changes, this is a breaking change.  I'm not sure how the Go SDK wants to handle breaking changes; it's still in preview so maybe you're not worried about them! 
 I think it's a breaking change worth taking too, but in case you disagree I'm happy to try to rewrite it to be a little more defensive on the API surface changes, I just didn't want to start there.

Separately, this is my first real change written in Go.  Please do not hesitate to call out any Go-isms and whatnot that I may have missed. :smile:

Resolves #168

### DIFF
--- a/azblob/zc_sas_query_params.go
+++ b/azblob/zc_sas_query_params.go
@@ -41,7 +41,7 @@ func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (st
 
 // SASTimeFormat represents the format of a SAS start or expiry time. Use it when formatting/parsing a time.Time.
 const SASTimeFormat = "2006-01-02T15:04:05Z" //"2017-07-27T00:00:00Z" // ISO 8601
-var SASTimeFormats = []string{SASTimeFormat, "2006-01-02T15:04Z", "2006-01-02"} // ISO 8601 formats, please refer to https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas for more details.
+var SASTimeFormats = []string{"2006-01-02T15:04:05.0000000Z", SASTimeFormat, "2006-01-02T15:04Z", "2006-01-02"} // ISO 8601 formats, please refer to https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas for more details.
 
 // formatSASTimeWithDefaultFormat format time with ISO 8601 in "yyyy-MM-ddTHH:mm:ssZ".
 func formatSASTimeWithDefaultFormat(t *time.Time) string {
--- a/azblob/zc_sas_query_params.go
+++ b/azblob/zc_sas_query_params.go
@@ -1,6 +1,7 @@
 package azblob
 
 import (
+	"errors"
 	"net"
 	"net/url"
 	"strings"
@@ -25,11 +26,11 @@ const (
 func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (string, string, string) {
 	ss := ""
 	if !startTime.IsZero() {
-		ss = startTime.Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
+		ss = formatSASTimeWithDefaultFormat(&startTime)
 	}
 	se := ""
 	if !expiryTime.IsZero() {
-		se = expiryTime.Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
+		se = formatSASTimeWithDefaultFormat(&expiryTime)
 	}
 	sh := ""
 	if !snapshotTime.IsZero() {
@@ -40,6 +41,37 @@ func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (st
 
 // SASTimeFormat represents the format of a SAS start or expiry time. Use it when formatting/parsing a time.Time.
 const SASTimeFormat = "2006-01-02T15:04:05Z" //"2017-07-27T00:00:00Z" // ISO 8601
+var SASTimeFormats = []string{SASTimeFormat, "2006-01-02T15:04Z", "2006-01-02"} // ISO 8601 formats, please refer to https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas for more details.
+
+// formatSASTimeWithDefaultFormat format time with ISO 8601 in "yyyy-MM-ddTHH:mm:ssZ".
+func formatSASTimeWithDefaultFormat(t *time.Time) string {
+	return formatSASTime(t, SASTimeFormat) // By default, "yyyy-MM-ddTHH:mm:ssZ" is used
+}
+
+// formatSASTime format time with given format, use ISO 8601 in "yyyy-MM-ddTHH:mm:ssZ" by default.
+func formatSASTime(t *time.Time, format string) string {
+	if format != "" {
+		return t.Format(format)
+	}
+	return t.Format(SASTimeFormat) // By default, "yyyy-MM-ddTHH:mm:ssZ" is used
+}
+
+// parseSASTimeString try to parse sas time string.
+func parseSASTimeString(val string) (t time.Time, timeFormat string, err error) {
+	for _, sasTimeFormat := range SASTimeFormats {
+		t, err = time.Parse(sasTimeFormat, val)
+		if err == nil {
+			timeFormat = sasTimeFormat
+			break
+		}
+	}
+
+	if err != nil {
+		err = errors.New("fail to parse time with IOS 8601 formats, please refer to https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas for more details")
+	}
+
+	return
+}
 
 // https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas
 
@@ -74,6 +106,10 @@ type SASQueryParameters struct {
 	signedExpiry       time.Time   `param:"ske"`
 	signedService      string      `param:"sks"`
 	signedVersion      string      `param:"skv"`
+
+	// private member used for startTime and expiryTime formatting.
+	stTimeFormat       string
+	seTimeFormat       string
 }
 
 func (p *SASQueryParameters) SignedOid() string {
@@ -202,9 +238,9 @@ func newSASQueryParameters(values url.Values, deleteSASParametersFromValues bool
 		case "snapshot":
 			p.snapshotTime, _ = time.Parse(SnapshotTimeFormat, val)
 		case "st":
-			p.startTime, _ = time.Parse(SASTimeFormat, val)
+			p.startTime, p.stTimeFormat, _ = parseSASTimeString(val)
 		case "se":
-			p.expiryTime, _ = time.Parse(SASTimeFormat, val)
+			p.expiryTime, p.seTimeFormat, _ = parseSASTimeString(val)
 		case "sip":
 			dashIndex := strings.Index(val, "-")
 			if dashIndex == -1 {
@@ -268,10 +304,10 @@ func (p *SASQueryParameters) addToValues(v url.Values) url.Values {
 		v.Add("spr", string(p.protocol))
 	}
 	if !p.startTime.IsZero() {
-		v.Add("st", p.startTime.Format(SASTimeFormat))
+		v.Add("st", formatSASTime(&(p.startTime), p.stTimeFormat))
 	}
 	if !p.expiryTime.IsZero() {
-		v.Add("se", p.expiryTime.Format(SASTimeFormat))
+		v.Add("se", formatSASTime(&(p.expiryTime), p.seTimeFormat))
 	}
 	if len(p.ipRange.Start) > 0 {
 		v.Add("sip", p.ipRange.String())
--- a/azblob/zt_url_blob_test.go
+++ b/azblob/zt_url_blob_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/md5"
 	"errors"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -1890,4 +1891,53 @@ func (s *aztestsSuite) TestBlobTierInvalidValue(c *chk.C) {
 
 	_, err = blobURL.SetTier(ctx, AccessTierType("garbage"), LeaseAccessConditions{})
 	validateStorageError(c, err, ServiceCodeInvalidHeaderValue)
+}
+
+func (s *aztestsSuite) TestBlobURLPartsSASQueryTimes(c *chk.C) {
+	StartTimesInputs := []string{
+		"2020-04-20",
+		"2020-04-20T07:00Z",
+		"2020-04-20T07:15:00Z",
+	}
+	StartTimesExpected := []time.Time{
+		time.Date(2020, time.April, 20, 0, 0, 0, 0, time.UTC),
+		time.Date(2020, time.April, 20, 7, 0, 0, 0, time.UTC),
+		time.Date(2020, time.April, 20, 7, 15, 0, 0, time.UTC),
+	}
+	ExpiryTimesInputs := []string{
+		"2020-04-21",
+		"2020-04-20T08:00Z",
+		"2020-04-20T08:15:00Z",
+	}
+	ExpiryTimesExpected := []time.Time{
+		time.Date(2020, time.April, 21, 0, 0, 0, 0, time.UTC),
+		time.Date(2020, time.April, 20, 8, 0, 0, 0, time.UTC),
+		time.Date(2020, time.April, 20, 8, 15, 0, 0, time.UTC),
+	}
+
+	for i := 0; i < len(StartTimesInputs); i++ {
+		urlString :=
+			"https://myaccount.blob.core.windows.net/mycontainer/mydirectory/myfile.txt?" +
+			"se=" + url.QueryEscape(ExpiryTimesInputs[i]) + "&" +
+			"sig=NotASignature&" +
+			"sp=r&" +
+			"spr=https&" +
+			"sr=b&" +
+			"st=" + url.QueryEscape(StartTimesInputs[i]) + "&" +
+			"sv=2019-10-10"
+		url, _ := url.Parse(urlString)
+
+		parts := NewBlobURLParts(*url)
+		c.Assert(parts.Scheme, chk.Equals, "https")
+		c.Assert(parts.Host, chk.Equals, "myaccount.blob.core.windows.net")
+		c.Assert(parts.ContainerName, chk.Equals, "mycontainer")
+		c.Assert(parts.BlobName, chk.Equals, "mydirectory/myfile.txt")
+
+		sas := parts.SAS
+		c.Assert(sas.StartTime(), chk.Equals, StartTimesExpected[i])
+		c.Assert(sas.ExpiryTime(), chk.Equals, ExpiryTimesExpected[i])
+
+		uResult := parts.URL()
+		c.Assert(uResult.String(), chk.Equals, urlString)
+	}
 }
--- a/azblob/zt_url_blob_test.go
+++ b/azblob/zt_url_blob_test.go
@@ -1898,21 +1898,25 @@ func (s *aztestsSuite) TestBlobURLPartsSASQueryTimes(c *chk.C) {
 		"2020-04-20",
 		"2020-04-20T07:00Z",
 		"2020-04-20T07:15:00Z",
+		"2020-04-20T07:30:00.1234567Z",
 	}
 	StartTimesExpected := []time.Time{
 		time.Date(2020, time.April, 20, 0, 0, 0, 0, time.UTC),
 		time.Date(2020, time.April, 20, 7, 0, 0, 0, time.UTC),
 		time.Date(2020, time.April, 20, 7, 15, 0, 0, time.UTC),
+		time.Date(2020, time.April, 20, 7, 30, 0, 123456700, time.UTC),
 	}
 	ExpiryTimesInputs := []string{
 		"2020-04-21",
 		"2020-04-20T08:00Z",
 		"2020-04-20T08:15:00Z",
+		"2020-04-20T08:30:00.2345678Z",
 	}
 	ExpiryTimesExpected := []time.Time{
 		time.Date(2020, time.April, 21, 0, 0, 0, 0, time.UTC),
 		time.Date(2020, time.April, 20, 8, 0, 0, 0, time.UTC),
 		time.Date(2020, time.April, 20, 8, 15, 0, 0, time.UTC),
+		time.Date(2020, time.April, 20, 8, 30, 0, 234567800, time.UTC),
 	}
 
 	for i := 0; i < len(StartTimesInputs); i++ {


AtOMiCNebula:
There are similar problems with SAS handling in `azure-storage-file-go` and `azure-storage-queue-go`, worth taking fixes over there too?

Also, I think CI tests might be broken; I see lots of failures when running against HEAD of `dev` locally, though that could be because there are some environment settings I don't have correctly in place. :innocent:

adreed-msft:
There is also a similar issue in the `azbfs` package of azcopy-- our integrated adlsg2 shim.

Much like my unexpectedEOF fixes, they're kind of everywhere.

AtOMiCNebula:
@adreed-msft I noticed the copies in azcopy too, I tried patching code in there first while trying to get azcopy to work, and found out that (at least for the SAS code) none of it was actually being executed, and that it was using the Go SDK's SAS code (hence this patchset).  Happy to get azcopy's fork patched too, though I don't know how I could test those changes if the code isn't getting reached.  Suggestions?

adreed-msft:
The code in there is specific to Azure Data Lakes Gen 2. You can use an existing blob sas against a dfs endpoint to test it.

AtOMiCNebula:
While copying some changes around, I found that the File SDK has some slight differences here, and has for over a year (Azure/azure-storage-file-go@85623f52f5b71fb0065aeee5c82833ee5da1a6ff).  It looks reasonably close to what I like to think I'd have put together for this to be a non-breaking-change (i.e. it adds additional `string` members, instead of switching `Time` to `string`).  It's not clear to me what value there is in storing the value as a `Time` at all, which is another reason I took the approach I did.

Not sure why this change was applied to the File SDK but not the other two.  But, thoughts?  For the File SDK, should I clean up this change to align the File SDK with this change completely, or should I just leave the File SDK alone (allowing it to keep a diverged set of routines).

azblob/sas_service.go
```diff
@@ -4,17 +4,16 @@ import (
...
 type BlobSASSignatureValues struct {
 	Version            string      `param:"sv"`  // If not specified, this defaults to SASVersion
 	Protocol           SASProtocol `param:"spr"` // See the SASProtocol* constants
-	StartTime          time.Time   `param:"st"`  // Not specified if IsZero
-	ExpiryTime         time.Time   `param:"se"`  // Not specified if IsZero
-	SnapshotTime       time.Time
+	StartTime          string      `param:"st"`
+	ExpiryTime         string      `param:"se"`
+	SnapshotTime       string
 	Permissions        string  `param:"sp"` // Create by initializing a ContainerSASPermissions or BlobSASPermissions and then call String()
```
adreed-msft, line 17: I am opposed to swapping time.Time out for strings here.

Why? Because when someone wants to create their SAS token and sign it here, they won't know what format to put in.

Personally, I'd prefer we have a pointer to a string for time format here. If it's nil, we should prefer our default time formatting for our SAS token. Then, at lower levels, we can either store the time.Time and the format OR we can store the string. I feel more comfortable with the former, how about you, @JohnRusk and @zezha-msft

azblob/zc_sas_account.go
```diff
@@ -5,16 +5,15 @@ import (
...
 // AccountSASSignatureValues is used to generate a Shared Access Signature (SAS) for an Azure Storage account.
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/constructing-an-account-sas
 type AccountSASSignatureValues struct {
 	Version       string      `param:"sv"`  // If not specified, this defaults to SASVersion
 	Protocol      SASProtocol `param:"spr"` // See the SASProtocol* constants
-	StartTime     time.Time   `param:"st"`  // Not specified if IsZero
-	ExpiryTime    time.Time   `param:"se"`  // Not specified if IsZero
+	StartTime     string      `param:"st"`
+	ExpiryTime    string      `param:"se"`
 	Permissions   string      `param:"sp"`  // Create by initializing a AccountSASPermissions and then call String()
```
adreed-msft, line 17: Also here

azblob/zc_sas_query_params.go
```diff
@@ -55,9 +36,9 @@ type SASQueryParameters struct {
 	services           string      `param:"ss"`
 	resourceTypes      string      `param:"srt"`
 	protocol           SASProtocol `param:"spr"`
-	startTime          time.Time   `param:"st"`
-	expiryTime         time.Time   `param:"se"`
-	snapshotTime       time.Time   `param:"snapshot"`
+	startTime          string      `param:"st"`
+	expiryTime         string      `param:"se"`
+	snapshotTime       string      `param:"snapshot"`
 	ipRange            IPRange     `param:"sip"`
```
adreed-msft, line 42: This is the "lower level" I was talking about.

azblob/zz_generated_models.go
```diff
@@ -6198,9 +6198,9 @@ type UserDelegationKey struct {
 	// SignedTid - The Azure Active Directory tenant ID in GUID format
 	SignedTid string `xml:"SignedTid"`
 	// SignedStart - The date-time the key is active
-	SignedStart time.Time `xml:"SignedStart"`
+	SignedStart string `xml:"SignedStart"`
 	// SignedExpiry - The date-time the key expires
```
adreed-msft, line 6202: If we modify a generated file like this, we should make it a swagger directive and regenerate the generated code.

azblob/parsing_urls.go
```diff
@@ -125,8 +125,8 @@ func (up BlobURLParts) URL() url.URL {
 	rawQuery := up.UnparsedParams
 
 	//If no snapshot is initially provided, fill it in from the SAS query properties to help the user
-	if up.Snapshot == "" && !up.SAS.snapshotTime.IsZero() {
-		up.Snapshot = up.SAS.snapshotTime.Format(SnapshotTimeFormat)
+	if up.Snapshot == "" && up.SAS.snapshotTime != "" {
+		up.Snapshot = up.SAS.snapshotTime
 	}
```
adreed-msft, line 130: I'm not certain snapshot time is affected by this @JohnRusk @zezha-msft

azblob/sas_service.go
JeffreyRichter: The reason for using Time is because these are times and code can easily take NOW add an hour (or whatever) to set the expiration time. Can the resolution problem be fixed by just changing the time format string: azblob.SASTimeFormat

azblob/zz_generated_models.go
AtOMiCNebula: I did not hand-edit this file, this change was the result of running swagger on my `swagger/blob.json` changes. :innocent:

azblob/sas_service.go
adreed-msft: I would say no. When stringifying the SAS token back out, we'd lose information as we currently do if we did that.

AtOMiCNebula: That's correct!  The problem is that the signature is generated with a specific string value, and if the SDK rewrites the expiryTime, then the service will not match the signature and fail the request.

To be clear, this affects existing SASs, not a SAS that the SDK itself creates from a given key.

azblob/zz_generated_models.go
adreed-msft: Yeah, that's also what I meant. Take a look at the files SDK. We don't have anything similar in the blob SDK because we've not had to do this yeat, but if we were to regenerate the blob SDK on a newer version, your change would go away. A proper swagger directive like defined below would allow it to stick between blob service versions.

https://github.com/Azure/azure-storage-file-go/tree/master/swagger

azblob/sas_service.go
JeffreyRichter: Can we get an example of a SAS time string that fails to convert to a Time due to resolution loss.
And, isn't this happening because of the format string? Why can we not change the format string to that resolution is maintained? Does Time not support the required resolution?

AtOMiCNebula: Changing the format string will fix my scenario, at the expense of breaking signatures that were generated with expirytimes without high precision, because now the problem is in reverse: times without high-precision information will be written out with it zero'd out (instead of none at all).

I'll get you an example broken URL in a bit.

azblob/zz_generated_models.go
AtOMiCNebula: Oh!!!  I did not realize blob.json _itself_ was also generated!  I'm less-new to Swagger than I am to Go, but still pretty new, and didn't know directives were a thing.  Will look into it, thank you for setting me straight. 😅

azblob/sas_service.go
JeffreyRichter: Here's another question? Why are you trying to round trip this? When I designed this, I expected most people will use the SDK to create SAS tokens. A lot fewer people might want to parse a SAS just to get the permissions, expiry time, etc. in order to some local processing. 

I never thought anyone would try to round trip the SAS since there really isn't any need to if you already have the original SAS. I'd hate to see the Time support go away in order to support a rare scenario. 

So, I'd like to understand your scenario more to see if it could be a more useful scenario that requires more attention.

adreed-msft: I don't think time support should go away, I just think we should keep track of the format to help round tripping. AzCopy for instance, cannot handle these "high precision" times in SAS tokens, due to the inability to round trip here.

JeffreyRichter: What is the round trip scenario? Why does AzCopy need to round trip SAS tokens?

adreed-msft: A URL with SAS token gets parsed by the blob SDK when attempting to enumerate, and the SAS token becomes invalid when attempting to make a request to the service, because the time format has changed.

adreed-msft: For another instance, the SAS token is separated from the URL with the blob SDK by clearing the SAS token values in a BlobURLParts struct, and then restringified down the line.

JeffreyRichter: This still isn't clear to me. Do you mean the customer passes a URL with a container or account SAS as a command-line argument. The SDK parses this using BlobURLParts? which has a SAS field with the parsed SAS and then AzCopy for some reason uses this SAS field to generate a new SAS and therefore the signature doesn't match?

If I got this right, then why do this? I guess you want the the SAS parameters parsed out of the URL for you easily, right? If so, why not just add a private string field to SASQueryParameters and have it hold the original SAS query parameters. If any file dis changed, invalid this field (set it to ""). If no field is changed, then have the function that gets a SAS token string return the field (if not "") and have it do what it does today if the field is "".

This way, nothing break; minimal code change, round-tripping now works. Times ALSO continue to work.

azblob/zz_generated_models.go
adreed-msft: No worries at all!

azblob/sas_service.go
adreed-msft: Yes, the intent was easy removal of the SAS token in that case.

I  don't feel like what you suggested (the private string field to the SAS query structure) is really a fix to the underlying problem here, more than it is sweeping it under the rug.

adreed-msft: There is still a loss of information going on in that case. The time itself does not have to be a string, but we should be aware of what time format succeeded when parsing the URL, and we should be aware of what time format the user wants in the generation structures, especially if there's a similar bug in another SDK or tool.

JeffreyRichter: It's not about SAS token removal because you want to re-apply it, right? I still can't say I full understand the whole flow and what's motivating this. I don't know why you think my proposal doesn't solve the problem or that it sweeps it under the rug. I'm trying to do minimal code change, no breaking changes, address the issue and keep the other features.

AtOMiCNebula: URL with SAS:
https://asbg169.blob.core.windows.net/example/hello.txt?sv=2015-04-05&sr=c&se=2020-04-14T17%3A00%3A00.0000000Z&sp=rl&sig=INgLizrwmX83P5J13gm03qZjCjAt5d5Xuqg%2BTrdSeqI%3D

In a browser, it works great.  AzCopy chokes, because somewhere along the way the SAS gets rewritten before it heads out to the network (note the error text listing an expiry time that is NOT what was passed in with the URL).

```
c:\...\azcopy_windows_amd64_10.4.0> ./azcopy.exe copy "https://asbg169.blob.core.windows.net/example/hello.txt?sv=2015-04-05&sr=c&se=2020-04-14T17%3A00%3A00.0000000Z&sp=rl&sig=INgLizrwmX83P5J13gm03qZjCjAt5d5Xuqg%2BTrdSeqI%3D" "hello.txt" --recursive
INFO: Scanning...
INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support

failed to perform copy command due to error: cannot start job due to error: cannot list blobs. Failed with error -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /home/vsts/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.7.0/azblob/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=AuthenticationFailed) =====
Description=Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:aaed7a13-c01e-0086-8084-0f5354000000
Time:2020-04-10T22:04:55.1065374Z, Details:
   AuthenticationErrorDetail: Signature did not match. String to sign used was rl

2020-04-14T17:00:00Z
/blob/asbg169/example



2015-04-05





   Code: AuthenticationFailed
   GET https://asbg169.blob.core.windows.net/example?comp=list&include=metadata&prefix=hello.txt%2F&restype=container&se=2020-04-14t17%3A00%3A00z&sig=-REDACTED-&sp=rl&sr=c&sv=2015-04-05&timeout=901
   User-Agent: [AzCopy/10.4.0 Azure-Storage/0.7 (go1.13; Windows_NT)]
   X-Ms-Client-Request-Id: [32777584-f87a-4757-6a49-deb8f8c2e1ff]
   X-Ms-Version: [2019-02-02]
   --------------------------------------------------------------------------------
   RESPONSE Status: 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
   Content-Length: [489]
   Content-Type: [application/xml]
   Date: [Fri, 10 Apr 2020 22:04:54 GMT]
   Server: [Microsoft-HTTPAPI/2.0]
   X-Ms-Error-Code: [AuthenticationFailed]
   X-Ms-Request-Id: [aaed7a13-c01e-0086-8084-0f5354000000]
```

AtOMiCNebula: (there is nothing sensitive in this blob, and the SAS will expire Tuesday @ Midnight Pacific time, lest anyone worry)

adreed-msft: What I'm suggesting isn't a breaking change-- Changing the time variables to strings is. Keeping them as times, and adding an additional variable to track the format actually used is not.

I say that your solution here sweeps it under the rug because, rather than actually tracking the time format, it just keeps the whole SAS token in addition to the parameters, and then regurgitates it when the user asks for the parameters to be stringified, rather than re-stringifying properly.

azblob/zc_sas_query_params.go
```diff
@@ -200,11 +181,11 @@ func newSASQueryParameters(values url.Values, deleteSASParametersFromValues bool
 		case "spr":
 			p.protocol = SASProtocol(val)
 		case "snapshot":
-			p.snapshotTime, _ = time.Parse(SnapshotTimeFormat, val)
+			p.snapshotTime = val
 		case "st":
```
adreed-msft, line 185: Just a note here, but in the Files SDK we can parse multiple time formats, I believe.

Something similar may be beneficial here.

azblob/sas_service.go
adreed-msft: > URL with SAS:
> [https://asbg169.blob.core.windows.net/example/hello.txt?sv=2015-04-05&sr=c&se=2020-04-14T17%3A00%3A00.0000000Z&sp=rl&sig=[REDACTED]](https://asbg169.blob.core.windows.net/example/hello.txt?sv=2015-04-05&sr=c&se=2020-04-14T17%3A00%3A00.0000000Z&sp=rl&sig=%5BREDACTED%5D)
> 
> In a browser, it works great. AzCopy chokes, because somewhere along the way the SAS gets rewritten before it heads out to the network (note the error text listing an expiry time that is NOT what was passed in with the URL).
> 
> ```
> c:\...\azcopy_windows_amd64_10.4.0> ./azcopy.exe copy "https://asbg169.blob.core.windows.net/example/hello.txt?sv=2015-04-05&sr=c&se=2020-04-14T17%3A00%3A00.0000000Z&sp=rl&sig=INgLizrwmX83P5J13gm03qZjCjAt5d5Xuqg%2BTrdSeqI%3D" "hello.txt" --recursive
> INFO: Scanning...
> INFO: Any empty folders will not be processed, because source and/or destination doesn't have full folder support
> 
> failed to perform copy command due to error: cannot start job due to error: cannot list blobs. Failed with error -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /home/vsts/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.7.0/azblob/zc_storage_error.go:42
> ===== RESPONSE ERROR (ServiceCode=AuthenticationFailed) =====
> Description=Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
> RequestId:aaed7a13-c01e-0086-8084-0f5354000000
> Time:2020-04-10T22:04:55.1065374Z, Details:
>    AuthenticationErrorDetail: Signature did not match. String to sign used was rl
> 
> 2020-04-14T17:00:00Z
> /blob/asbg169/example
> 
> 
> 
> 2015-04-05
> 
> 
> 
> 
> 
>    Code: AuthenticationFailed
>    GET https://asbg169.blob.core.windows.net/example?comp=list&include=metadata&prefix=hello.txt%2F&restype=container&se=2020-04-14t17%3A00%3A00z&sig=-REDACTED-&sp=rl&sr=c&sv=2015-04-05&timeout=901
>    User-Agent: [AzCopy/10.4.0 Azure-Storage/0.7 (go1.13; Windows_NT)]
>    X-Ms-Client-Request-Id: [32777584-f87a-4757-6a49-deb8f8c2e1ff]
>    X-Ms-Version: [2019-02-02]
>    --------------------------------------------------------------------------------
>    RESPONSE Status: 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
>    Content-Length: [489]
>    Content-Type: [application/xml]
>    Date: [Fri, 10 Apr 2020 22:04:54 GMT]
>    Server: [Microsoft-HTTPAPI/2.0]
>    X-Ms-Error-Code: [AuthenticationFailed]
>    X-Ms-Request-Id: [aaed7a13-c01e-0086-8084-0f5354000000]
> ```

I censored it out still because anyone who gets ahold of the SAS token could make a simple query hundreds of thousands of times and cost you money.

AtOMiCNebula: I like to live dangerously (and wanted to provide a short-lived repro), but hopefully what remains in the paste is sufficient for understanding of the issue.  Thanks for being a cooler-head for me.

To try to start tying off a long-thread, I started out thinking there wasn't a big need to keep the `Time` around, but after sleeping on it a bit agree that it would be nice to keep.  I'm going to look a little closer at the File SDK change I called out earlier, and just port that around if it's sufficient for this issue, and will fix and then port if it needs additional work (it doesn't contain the high-precision time format, though I don't know if it needs it).  I still don't like that it tries to do any parsing of the string, but I'm fine with that being a me-thing. :smile:

azblob/zc_sas_query_params.go
AtOMiCNebula: Yes, I called that change out earlier today I think.  I still don't like that it's trying to parse a `Time` out from the string, even if using multiple format strings, because even if we retain the string and send it back out, the `Time` value itself may not match and could lead to confusion.  But I see the value in letting the API maintain a `Time` too.

AtOMiCNebula:
Second iteration is a near-port of Azure/azure-storage-file-go@85623f5 and Azure/azure-storage-file-go@edcebb6 (plus the addition of the high-precision time format).  I largely rewrote the test, since it didn't port cleanly (File SASs are different enough, and it was a lot of copy-pasta that was different in subtle ways that I've tried to clean up).

I still don't like this approach, because it requires specific timestamp patterns to be specified in code, so I would not be surprised if this list needs further iteration in the future as people use their own timestamp formats that the Azure service accepts (but the Go SDK mangles).  There's a more thorough fix here that is a combination of my original attempt and this one, but I don't have time to assemble it. ☹️

adreed-msft:
> I still don't like this approach, because it requires specific timestamp patterns to be specified in code, so I would not be surprised if this list needs further iteration in the future as people use their own timestamp formats that the Azure service accepts (but the Go SDK mangles). There's a more thorough fix here that is a combination of my original attempt and this one, but I don't have time to assemble it. ☹️

It seems like a future item to tackle that we infer the actual timestamp format from the text.

ISO-8601 has a very specific ordering and format, so I wonder how difficult that'd be.

AtOMiCNebula:
@adreed-msft / @JeffreyRichter I resolved some merge conflicts, please take a look at this revised version (and/or let me know who else to ping :smile:

adreed-msft:
Ah, pardon the slow response here. I'll try and get to reviewing this today.

adreed-msft:
Twenty-six days ago, sorry! Taking a deep look now-- thanks for your work.

azblob/zc_sas_query_params.go
```diff
@@ -40,6 +41,37 @@ func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (st
 
 // SASTimeFormat represents the format of a SAS start or expiry time. Use it when formatting/parsing a time.Time.
 const SASTimeFormat = "2006-01-02T15:04:05Z" //"2017-07-27T00:00:00Z" // ISO 8601
+var SASTimeFormats = []string{"2006-01-02T15:04:05.0000000Z", SASTimeFormat, "2006-01-02T15:04Z", "2006-01-02"} // ISO 8601 formats, please refer to https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas for more details.
+
+// formatSASTimeWithDefaultFormat format time with ISO 8601 in "yyyy-MM-ddTHH:mm:ssZ".
+func formatSASTimeWithDefaultFormat(t *time.Time) string {
+	return formatSASTime(t, SASTimeFormats[1]) // By default, "yyyy-MM-ddTHH:mm:ssZ" is used
```
adreed-msft, line 48: >SASTimeFormats[1] [](start = 25, length = 17)

Since we're already using a constant outside of this list, let's avoid magic numbers by referencing it directly (SASTimeFormat), and perhaps rename it to DefaultSASTimeFormat.

azblob/zc_sas_query_params.go
```diff
@@ -40,6 +41,37 @@ func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (st
...
+}
+
+// formatSASTime format time with given format, use ISO 8601 in "yyyy-MM-ddTHH:mm:ssZ" by default.
+func formatSASTime(t *time.Time, format string) string {
+	if format != "" {
+		return t.Format(format)
+	}
+	return t.Format(SASTimeFormats[1]) // By default, "yyyy-MM-ddTHH:mm:ssZ" is used
```
adreed-msft, line 56: >SASTimeFormats[1] [](start = 17, length = 17)

Ditto

azblob/zc_sas_query_params.go
```diff
@@ -40,6 +41,37 @@ func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (st
...
+var SASTimeFormats = []string{"2006-01-02T15:04:05.0000000Z", SASTimeFormat, "2006-01-02T15:04Z", "2006-01-02"} // ISO 8601 formats, please refer to https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas for more details.
+
+// formatSASTimeWithDefaultFormat format time with ISO 8601 in "yyyy-MM-ddTHH:mm:ssZ".
+func formatSASTimeWithDefaultFormat(t *time.Time) string {
+	return formatSASTime(t, SASTimeFormats[1]) // By default, "yyyy-MM-ddTHH:mm:ssZ" is used
+}
+
+// formatSASTime format time with given format, use ISO 8601 in "yyyy-MM-ddTHH:mm:ssZ" by default.
+func formatSASTime(t *time.Time, format string) string {
+	if format != "" {
```
adreed-msft, line 53: These aren't the only places that a SAS token gets parsed or formatted.

Check where SASTimeFormat is referenced, and potentially the time formats list.

adreed-msft: For instance, zc_sas_query_params.go:205 to 207

---
In reply to: [437652454](https://github.com/Azure/azure-storage-blob-go/pull/169#discussion_r437652454) [](ancestors = 437652454)

azblob/zt_url_blob_test.go
```diff
@@ -1891,3 +1892,56 @@ func (s *aztestsSuite) TestBlobTierInvalidValue(c *chk.C) {
 	_, err = blobURL.SetTier(ctx, AccessTierType("garbage"), LeaseAccessConditions{})
 	validateStorageError(c, err, ServiceCodeInvalidHeaderValue)
 }
+
+func (s *aztestsSuite) TestBlobURLPartsSASQueryTimes(c *chk.C) {
+	StartTimesInputs := []string{
```
adreed-msft, line 1897: Nice test!

azblob/zc_sas_query_params.go
AtOMiCNebula: Ha!! Great suggestion on referring to the constant directly, wish I had realized this myself. 😇 

I don't want to rename `SASTimeFormat` itself though, since you already convinced me to not break the API and this would be a breaking change (https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#pkg-constants).

azblob/zt_url_blob_test.go
AtOMiCNebula: I don't know if you said this realizing my test wasn't compiling or not, but I must have overlooked that somehow when I originally posted this for review...golang continues to get the best of me. 😦

adreed-msft: Huh, it's compiling for me-- just not passing. I couldn't figure out quite why, because it would run fine to the end in a debugger-- But it wouldn't return a result.

azblob/zc_sas_query_params.go
adreed-msft: Good thought, thanks for the catch.

azblob/zt_url_blob_test.go
AtOMiCNebula: Genuinely no clue how I didn't spot this error while writing the test.  Either way, it is now fixed.

adreed-msft:
By the way, I'd ported your work to the blobfs SDK and updated the files SDK for it in secondary branches to create a custom build of azcopy for internal usage-- Works like a charm!

AtOMiCNebula:
@adreed-msft Tests still seem to be failing in CI, lots of them.  I don't think the failures are related to my change, and I can get panics locally when running some of the affected tests without my changes applied.  Is the test suite flaky, or am I just missing something obvious? 😕

AtOMiCNebula:
Based on https://travis-ci.org/github/Azure/azure-storage-blob-go/pull_requests, it looks like all PR validation builds are failing, though normal CI validation looks okay (https://travis-ci.org/github/Azure/azure-storage-blob-go/branches).

zezha-msft:
Hi @AtOMiCNebula, no worries, the CI fails because the test credentials are only available if the PR originates from inside the repo. I'll merge and it should run just fine.

PR #171, opened 2020-04-17
Convert SAS times to UTC before formatting

### LINKED ISSUE #170: FormatTimesForSASSigning should convert to UTC before formatting
v0.8.0
Arch Linux
Passing in `time.Now().Add(time.Hour)` to `BlobSASSignatureValues.ExpiryTime` results in the time string dropping TZ data and assuming it's in UTC. My local is in EST but the format string omits any Time Zone addition with a static `Z` ending: `2006-01-02T15:04:05Z`
https://play.golang.org/p/YB02MQG3m1T

Something like
```go
azblob.FormatTimesForSASSigning(time.Now().In(time.FixedZone("UTC-8", -8*60*60), time.Time{}, time.TIme{})
```

It should be an easy fix - could have saved me a load of hours as I saw the time format with a `Z` ending in the encoded query and assumed it was properly converting to UTC (didn't pay close enough attention to the actual numbers though)

Pass in the time values in the struct as `UTC` myself

someone1:
Fixes #170

### DIFF
--- a/azblob/zc_sas_query_params.go
+++ b/azblob/zc_sas_query_params.go
@@ -25,15 +25,15 @@ const (
 func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (string, string, string) {
 	ss := ""
 	if !startTime.IsZero() {
-		ss = startTime.Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
+		ss = startTime.UTC().Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
 	}
 	se := ""
 	if !expiryTime.IsZero() {
-		se = expiryTime.Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
+		se = expiryTime.UTC().Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
 	}
 	sh := ""
 	if !snapshotTime.IsZero() {
-		sh = snapshotTime.Format(SnapshotTimeFormat)
+		sh = snapshotTime.UTC().Format(SnapshotTimeFormat)
 	}
 	return ss, se, sh
 }


zezha-msft:
Hi @someone1, thanks for the PR!

Could you please address the comment from @AtOMiCNebula?

someone1:
@zezha-msft - I can certainly try!

@AtOMiCNebula - I don't know if I follow as I'm not thoroughly familiar with the process you're describing. Do you mind elaborating a bit? Seeing #169 - I'm inclined to just add `t = t.UTC()` to the start of the `formatSASTime` function, and also use it for the snapshot time `sh` in `FormatTimesForSASSigning`. Would that be sufficient?

The timestamps seem always to be assumed to be in UTC in the formats - so I'm not sure how you you can generate valid signatures with non-UTC timestamps as you've stated.

AtOMiCNebula:
@someone1 This library has a bad habit of rewriting components of SASs, which breaks the ability to actually _use_ the SAS.  The fix of mine that I referred to earlier was to address an issue where AzCopy was unable to use a SAS produced by Azure itself, because Azure was handing out SASs with timestamps that included a microseconds component.  This SDK stripped out the microseconds, and you can't do that, since the signature is invalidated (by design) if you change any of the parts.

My concern is that this change may break round-tripping (though perhaps I came off a bit strong in my earlier statement).  The scenario I have in mind is that a string timestamp is parsed into a `Time` instance that is _not_ UTC, since your change will cause it to be serialized/formatted out to a materially different string than it was before (and thus break the SAS).  For instance, [from Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC), `"2020-06-24T21:13:05-0800"`.  However, upon further reflection, the buggy processing may strip off the time zone identifier in the first place (I haven't tested it), so maybe this whole concern is moot.  Either way, please ensure you do not break this in the event that it _does_ presently work. :smile:

azblob/zc_sas_query_params.go
```diff
@@ -25,15 +25,15 @@ const (
 func FormatTimesForSASSigning(startTime, expiryTime, snapshotTime time.Time) (string, string, string) {
 	ss := ""
 	if !startTime.IsZero() {
-		ss = startTime.Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
+		ss = startTime.UTC().Format(SASTimeFormat) // "yyyy-MM-ddTHH:mm:ssZ"
```
mohsha-msft, line 28: Hey @someone1 ,

Thanks for reaching out !
We are happy to see your contribution.

There are couple of things which I'd like to point out
1. We always merge to `dev` branch before releasing it to master. So I changed the base to `dev` branch
2. There is a conflict in merge which will be removed if you do the required changes in either `formatSASTimeWithDefaultFormat()` function or send UTC time as argument itself.

Also, it'd be nice if you can see whether `formatSASTimeWithDefaultFormat()` func is getting used some other function call because if that is the case, we need to see if UTC time is required in that function or not?

PR #172, opened 2020-05-19
Set `Parallelism` default before using it

kfarnung:
The `Parallelism` value isn't set to the default value until after the `operationChannel` is created. This causes the `operationChannel` to be created with a buffer size of zero (unbuffered) instead of what I assume to be the intended value of `5`. This causes the Go routines to bottleneck on the channel as only one message can pass through at a time.


kfarnung:
CI run: https://travis-ci.com/github/kfarnung/azure-storage-blob-go/builds/166837023

zezha-msft:
Hi @kfarnung, thanks so much for the PRs! And I really apologize for the delay in reviewing them. We'll go through them shortly.

PR #174, opened 2020-05-26
[chunkwriting] Handle data races in chunk IDs

kfarnung:
During test execution `TestCopyFromReader` hit a data race which
resulted in the test failing due to checksum mismatch.

Since the `id` struct's methods are not thread-safe there is a race
condition where two upload threads request an ID at the same moment and
both could potentially end up with the same ID.

The simplest solution is to move the ID assignment to the thread that
populates the chunks to the channel. This ensures that the chunk IDs are
assigned sequentially and removes the race condition.

Introduced by #167

PR #175, opened 2020-05-26
[highlevel] Stop using memory-mapped files

### LINKED ISSUE #119: [Proposal] High-level APIs should support to taking io.WriterAt as output
Latest

Linux

No support for io.writerAt.

N/A

Mitigate by writing directly to a file.

### Description
I am writing a plugin for a parallel file system to archive back to BLOB storage.  The current implementation handles the IO but requires io.reader for upload and io.writerAt for download.  There is currently no support for io.writerAt.  Would it be possible to add?

kfarnung:
While investigating this SDK for uploading and downloading large blobs
(e.g. 25GB or more) it became apparent that the memory-mapped approach
has some severe limitations:

1. Limits the file size on 32-bit systems (theoretically 4GB, but much
   less in practice).
2. Has no backpressure when writing to slower storage mediums.
3. Appears to finish faster, but the OS spends several minutes flushing
   the modified RAM to disk afterwards (depends on the speed of the
   disk).

On a VM with 16GB of RAM and a slow disk (spinning in this case) the
algorithm quickly overwhelms the available memory and causes severe
performance degradation. It ended up simultaneously trying to flush to
the slow data disk and page out to the slightly faster OS disk.

The solution is to stop using memory-mapped files (at least the way the
SDK currently uses then) and switch to the `io.ReaderAt` and
`io.WriterAt` interfaces. They explicitly allow for parallel access to
non-overlapping regions which make them a good candidate for this
purpose.

Benchmarking large downloads (25GB file) between azcopy 10.4.3 and these
updates using a test app, the difference between them is within 10
seconds. When compared against the original code on a beefy machine with
plenty of RAM the measured execution time is faster, but there is a
little bit of delay while the last of the data flushes from RAM to disk.

Fixes #119


kfarnung:
CI run: https://travis-ci.com/github/kfarnung/azure-storage-blob-go/builds/168151969

FYI @element-of-surprise, This change always seems to hit the race condition I attempt to fix in #174. I can't really explain why other than maybe the timing changes with more tests running in parallel?

kfarnung:
I merged the changes from #174 into this PR and ran a test build:

https://travis-ci.com/github/kfarnung/azure-storage-blob-go/builds/168238361

Looks like the race detection is gone.

zezha-msft:
Thanks @kfarnung for the contribution!

We'll have a discussion on our end and get back to you shortly.

kfarnung:
Absolutely understand. I was playing around to see how feasible it would be and figured a PR might explain better than words.

In the meantime I’ve created a partial fork for my own usage:

https://github.com/kfarnung/go-azblobext

Let me know what you decide, if this lands I can discontinue using my forked code.

azblob/section_writer_test.go
```diff
@@ -0,0 +1,85 @@
...
+	c.Assert(section.position, chk.Equals, int64(5))
+	c.Assert(b, chk.Equals, [10]byte{1, 2, 3, 4, 5, 0, 0, 0, 0, 0})
+
+	count, err = section.Write([]byte{6, 7, 8})
+	c.Assert(err, chk.ErrorMatches, "End of section reached")
+	c.Assert(count, chk.Equals, 0)
+	c.Assert(section.position, chk.Equals, int64(5))
+	c.Assert(b, chk.Equals, [10]byte{1, 2, 3, 4, 5, 0, 0, 0, 0, 0})
+
+	section = newSectionWriter(buffer, 5, 6)
```
zezha-msft, line 37: >6 [](start = 39, length = 1)

the 6 is on purpose right? there are only 5 bytes worth of space, but even if we set it to 6, the underlying bytes writer will only write 5 bytes in total. A comment would be helpful here.

kfarnung: Good point, thanks! I've added some comments to help clarify this section.

nakulkar-msft (approved):

Issue #173, opened 2020-05-25
go.mod doesn't track all project's dependencies


simonpasquier:
### Which version of the SDK was used?
v0.8.0

### Which platform are you using? (ex: Windows, Linux, Debian)
Linux

### What problem was encountered?

The current `go.mod` doesn't track all project's dependencies which can cause issues for project using this library.

### How can we reproduce the problem in the simplest way?

Run `go mod tidy && git diff` on a fresh cloned copy of the project.

### Have you found a mitigation/solution?
No

zezha-msft:
Sorry for the inconvenience @simonpasquier. This should be fixed in 0.9 now.

simonpasquier:
Thanks a lot!

PR #177, opened 2020-06-03
Update dependencies

jhendrixMSFT:
Some dependencies are missing from the module leading to ambiguous
import errors in upstream projects.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-storage-blob-go/issues/173

PR #188, opened 2020-07-10
Blob versioning

mohsha-msft:


azblob/sas_service.go
```diff
@@ -44,6 +44,14 @@ func (v BlobSASSignatureValues) NewSASQueryParameters(credential StorageAccountC
 			return SASQueryParameters{}, err
 		}
 		v.Permissions = perms.String()
+	} else if v.Version != null && v.Version != "" {
+		resource = "bv"
+		//Make sure the permission characters are in the correct order
+		perms := &BlobSASPermissions{}
+		if err := perms.Parse(v.Permissions); err != nil {
+			return SASQueryParameters{}, err
+		}
```
zezha-msft, line 53: testing needed

azblob/url_blob.go
```diff
@@ -108,16 +120,22 @@ func (b BlobURL) Undelete(ctx context.Context) (*BlobUndeleteResponse, error) {
 // does not update the blob's ETag.
 // For detailed information about block blob level tiering see https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-storage-tiers.
 func (b BlobURL) SetTier(ctx context.Context, tier AccessTierType, lac LeaseAccessConditions) (*BlobSetTierResponse, error) {
-	return b.blobClient.SetTier(ctx, tier, nil, RehydratePriorityNone, nil, lac.pointers())
+	return b.blobClient.SetTier(ctx, tier, nil,
+		nil, // Blob versioning
```
zezha-msft, line 124: >// Blob versioning [](start = 7, length = 18)

please change these comments to clarify that the blob version is already included in the URL

zezha-msft:
We need some basic testing for the version functionality, the same coverage as the Python PR would be nice.

azblob/url_blob.go
mohsha-msft: Done

PR #190, opened 2020-07-15
#7508079 [Go][Blob][2019-12-12] Blob Versioning

mohsha-msft:
Blob versioning feature provides overwrite/delete protection to a blob by automatically generating a snapshot whenever the user makes a change or delete a blob. The automatically generated snapshot timestamp (version-id) is returned to user.
Please read [this](https://docs.microsoft.com/en-us/azure/storage/blobs/versioning-overview?tabs=powershell) for more information on blob versioning.

Note: Blob versioning (preview) is currently enabled in few regions only.


azblob/parsing_urls.go
```diff
@@ -124,6 +134,11 @@ func (up BlobURLParts) URL() url.URL {
 
 	rawQuery := up.UnparsedParams
 
+	// Check: Both snapshot and version id cannot be present in the request URL.
+	if up.Snapshot != "" && up.VersionID != "" {
+		log.Fatal("Snapshot and versioning cannot be enabled simultaneously.")
```
zezha-msft, line 139: >log.Fatal("Snapshot and versioning cannot be enabled simultaneously.") [](start = 2, length = 70)

We cannot do fatal logging in any SDK: // Fatal is equivalent to Print() followed by a call to os.Exit(1).

Let's avoid this check, or return an error

azblob/zt_url_blob_test.go
```diff
@@ -1737,23 +1737,23 @@ func (s *aztestsSuite) TestBlobSetMetadataIfNoneMatchFalse(c *chk.C) {
...
+	//_, err = blobURL.Undelete(ctx)
+	//if err != nil { // We want to give the wrapper method a chance to check if it was an error related to the service properties update.
+	//	return err
+	//}
+	//
+	//resp, err := blobURL.GetProperties(ctx, BlobAccessConditions{})
+	//if err != nil {
+	//	return errors.New(string(err.(StorageError).ServiceCode()))
+	//}
+	//c.Assert(resp.BlobType(), chk.Equals, BlobBlockBlob) // We could check any property. This is just to double check it was undeleted.
```
zezha-msft, line 1756: does turning on versioning disable soft delete?

azblob/zt_url_blob_test.go
```diff
@@ -1970,3 +1970,192 @@ func (s *aztestsSuite) TestDownloadBlockBlobUnexpectedEOF(c *chk.C) {
...
+
+	// Deleting previous version snapshot.
+	deleteResp, err := blobURL.WithVersionID(versionId1).Delete(ctx, DeleteSnapshotsOptionNone, BlobAccessConditions{})
+	c.Assert(err, chk.IsNil)
+	c.Assert(deleteResp.StatusCode(), chk.Equals, 202)
+
+	listBlobsResp, err = containerURL.ListBlobsFlatSegment(ctx, Marker{}, ListBlobsSegmentOptions{Details: BlobListingDetails{Versions: true}})
+	c.Assert(err, chk.IsNil)
+	c.Assert(listBlobsResp.Segment.BlobItems, chk.NotNil)
+	if len(listBlobsResp.Segment.BlobItems) == 0 {
```
zezha-msft, line 2044: >len(listBlobsResp.Segment.BlobItems) == 0 [](start = 4, length = 41)

can we be more precise with the validation? Instead of just making sure it's not 0, make sure that it's actually 1

azblob/zt_url_blob_test.go
```diff
@@ -1970,3 +1970,192 @@ func (s *aztestsSuite) TestDownloadBlockBlobUnexpectedEOF(c *chk.C) {
...
+		c.Fail()
+	}
+}
+
+func (s *aztestsSuite) TestDeleteSpecificBlobVersionWithBlobSAS(c *chk.C) {
```
zezha-msft, line 2049: it might be a good idea to isolate the versioning tests into a separate file

azblob/zt_url_blob_test.go
```diff
@@ -1970,3 +1970,192 @@ func (s *aztestsSuite) TestDownloadBlockBlobUnexpectedEOF(c *chk.C) {
...
+		c.Fatal(err)
+	}
+
+	sbURL := NewBlockBlobURL(blobParts.URL(), containerURL.client.p)
+	deleteResp, err := sbURL.Delete(ctx, DeleteSnapshotsOptionNone, BlobAccessConditions{})
+	c.Assert(deleteResp, chk.IsNil)
+
+	listBlobResp, err := containerURL.ListBlobsFlatSegment(ctx, Marker{}, ListBlobsSegmentOptions{Details: BlobListingDetails{Versions: true}})
+	c.Assert(err, chk.IsNil)
+	c.Assert(listBlobResp.Version(), chk.Equals, ServiceVersion)
```
zezha-msft, line 2089: >c.Assert(listBlobResp.Version(), chk.Equals, ServiceVersion) [](start = 1, length = 60)

a bit confused by this validation, why are we validating the service version?

azblob/zt_url_blob_test.go
zezha-msft: I see, "If both versioning and soft delete are enabled on a storage account, then no soft-deleted snapshot is created when a blob or blob version is modified or deleted."

---
In reply to: [457807890](https://github.com/Azure/azure-storage-blob-go/pull/190#discussion_r457807890) [](ancestors = 457807890)

azblob/zt_url_block_blob_test.go
```diff
@@ -898,3 +899,149 @@ func (s *aztestsSuite) TestBlobPutBlockListModifyBlob(c *chk.C) {
...
+	blobURL := containerURL.NewBlockBlobURL(generateBlobName())
+
+	data := []string{"Azure ", "Storage ", "Block ", "Blob."}
+	base64BlockIDs := make([]string, len(data))
+
+	for index, d := range data {
+		base64BlockIDs[index] = blockIDIntToBase64(index)
+		resp, err := blobURL.StageBlock(ctx, base64BlockIDs[index], strings.NewReader(d), LeaseAccessConditions{}, nil)
+		if err != nil {
+			log.Fatal(err)
```
zezha-msft, line 1029: >log.Fatal(err) [](start = 3, length = 14)

let's not use log.fatal, we can use c.Fail instead

zezha-msft:
> 

CopyFromURL is already sync

---
Refers to: azblob/zt_url_block_blob_test.go:1048 in 08946a3. [](commit_id = 08946a34d625246a5e315e40ecbd72b811d0a445, deletion_comment = False)

azblob/parsing_urls.go
mohsha-msft: Ohh I see. Alright. I'll do `errors.New()` to throw this error.

azblob/zt_url_blob_test.go
mohsha-msft: Yes exactly. When both versioning and soft-delete features are enabled, then soft-delete snapshot won't be created. That's why I commented this test case out for now.

azblob/zt_url_blob_test.go
mohsha-msft: Done.

azblob/zt_url_block_blob_test.go
mohsha-msft: Done.

azblob/zt_url_blob_test.go
mohsha-msft: Versioning is supported by service version `2019-12-12` or higher. I used this validation as a way to check/flag that versioning is enabled on the storage account.

azblob/zt_url_blob_test.go
mohsha-msft: So, are you referring to all the test cases or just select few? Yeah I think this is a good idea because we can avoid making test case files large and monolithic but if we follow this practice and bifurcate tests in their separate files for all the features, #files will become more eventually.

mohsha-msft: Done

azblob/parsing_urls.go
```diff
@@ -124,6 +134,11 @@ func (up BlobURLParts) URL() url.URL {
 
 	rawQuery := up.UnparsedParams
 
+	// Check: Both snapshot and version id cannot be present in the request URL.
+	if up.Snapshot != "" && up.VersionID != "" {
+		errors.New("Snapshot and versioning cannot be enabled simultaneously")
```
zezha-msft, line 139: >errors.New("Snapshot and versioning cannot be enabled simultaneously") [](start = 2, length = 70)

This is not getting returned though? you can also avoid this check and simply let the service give back an error message

mohsha-msft: Ohh understood. I'll remove.  Since blob versioning and snapshot cannot be used at the same time I was thinking of putting a check in our code base.

azblob/zt_blob_versioning_test.go
```diff
@@ -0,0 +1,386 @@
...
+	deleteResp, err := blobURL.WithVersionID(versionID1).Delete(ctx, DeleteSnapshotsOptionNone, BlobAccessConditions{})
+	c.Assert(err, chk.IsNil)
+	c.Assert(deleteResp.StatusCode(), chk.Equals, 202)
+
+	listBlobsResp, err = containerURL.ListBlobsFlatSegment(ctx, Marker{}, ListBlobsSegmentOptions{Details: BlobListingDetails{Versions: true}})
+	c.Assert(err, chk.IsNil)
+	c.Assert(listBlobsResp.Segment.BlobItems, chk.NotNil)
+	if len(listBlobsResp.Segment.BlobItems) != 1 {
+		c.Fail()
+	}
```
zezha-msft, line 107: If there's a single item, why don't we check to make sure that the right version remains?

azblob/zt_blob_versioning_test.go
```diff
@@ -0,0 +1,386 @@
...
+	c.Assert(deleteResp.Response().Header.Get("x-ms-version-id"), chk.NotNil)
+
+	listBlobResp, err = containerURL.ListBlobsFlatSegment(ctx, Marker{}, ListBlobsSegmentOptions{Details: BlobListingDetails{Snapshots: true, Versions: true}})
+	c.Assert(err, chk.IsNil)
+	c.Assert(listBlobResp.rawResponse.Header.Get("x-ms-request-id"), chk.NotNil)
+	if len(listBlobResp.Segment.BlobItems) == 0 {
+		c.Fail()
+	}
+	blobs := listBlobResp.Segment.BlobItems
+	c.Assert(blobs[0].Snapshot, chk.Equals, "")
```
zezha-msft, line 327: Can we be more precise with the test validation?

azblob/zt_blob_versioning_test.go
```diff
@@ -0,0 +1,386 @@
...
+
+	contentResp, err := blobURL.Download(ctx, 0, CountToEnd, BlobAccessConditions{}, false)
+	c.Assert(err, chk.IsNil)
+	contentData, err := ioutil.ReadAll(contentResp.Body(RetryReaderOptions{}))
+	c.Assert(contentData, chk.DeepEquals, []uint8(strings.Join(data, "")))
+}
+
+func (s *aztestsSuite) TestSyncCopyBlobReturnsVID(c *chk.C) {
+
+}
```
zezha-msft, line 367: empty?

azblob/zt_url_blob_test.go
zezha-msft: this will always succeed thought, right? Since the SDK will use the Service Version by default

PR #194, opened 2020-07-17
Add illumos build tag additionally to solaris

Toasterson:
This is more to future proof support for both systems. The illumos build tag ATM inherits everything from the solaris build tag but may not in the future.  It's generally best practice to add both build tags.

@petems FYI


zezha-msft:
Thanks @Toasterson for the contribution!

PR #200, opened 2020-08-11
Minor versioning fix

mohsha-msft:


azblob/parsing_urls.go
```diff
@@ -99,7 +99,8 @@ func NewBlobURLParts(u url.URL) BlobURLParts {
 	if versionIDs, ok := caseInsensitiveValues(paramsMap).Get(versionId); ok {
 		up.VersionID = versionIDs[0]
 		// If we recognized the query parameter, remove it from the map
-		delete(paramsMap, versionId)
+		delete(paramsMap, versionId)   // delete "versionid" from paramsMap
+		delete(paramsMap, "versionId") // delete "versionId" from paramsMap
```
zezha-msft, line 103: Minor: maybe add a comment here to clarify why there's such a case.

mohsha-msft: Alright. I'll do that and push that along with the set tier task I am doing. But if you look closely, the code snippet from line 93-104 are incorrect. When the code looks for a parameter in the paramsMap, it actually looks for case-insensitive key. However when code tries to delete the same parameter from paramsMap, it deleted actual case-sensitive key.

PR #203, opened 2020-08-24
[Go][Blob][2019-02-02] Set tier support on copy/put blob API

mohsha-msft:
Instead of invoking `setTier() `method separately, added support to set tier of  blob at the time of upload/copy. This will facilitate faster performance. 

Changed the function signature of `Upload()` and introduced new parameter tier of type `AccessTierType` .

### DIFF
--- a/azblob/zt_url_block_blob_test.go
+++ b/azblob/zt_url_block_blob_test.go
@@ -939,7 +939,7 @@ func (s *aztestsSuite) TestSetTierOnCopyBlockBlobFromURL(c *chk.C) {
 	bsu := getBSU()
 
 	container, _ := createNewContainer(c, bsu)
-	defer delContainer(c, container)
+	//defer delContainer(c, container)
 
 	testSize := 1 * 1024 * 1024
 	r, sourceData := getRandomDataAndReader(testSize)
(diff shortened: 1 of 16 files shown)


azblob/zt_url_block_blob_test.go
```diff
@@ -899,3 +899,88 @@ func (s *aztestsSuite) TestBlobPutBlockListModifyBlob(c *chk.C) {
...
+		_, err = blobURL.CommitBlockList(ctx, []string{blockID}, BlobHTTPHeaders{}, nil, BlobAccessConditions{}, tier)
+
+		resp, err := blobURL.GetBlockList(ctx, BlockListCommitted, LeaseAccessConditions{})
+		c.Assert(err, chk.IsNil)
+		c.Assert(resp.CommittedBlocks, chk.HasLen, 1)
+		c.Assert(resp.UncommittedBlocks, chk.HasLen, 0)
+	}
+}
+
+func (s *aztestsSuite) TestSetTierOnCopyBlockBlobFromURL(c *chk.C) {
```
zezha-msft, line 938: maybe I missed it, but we need a test for the async copy method as well

mohsha-msft: Sure. Will do that 👍

PR #204, opened 2020-08-24
Consider 502 as a temporary error

zezha-msft:
From MDN:

> The HyperText Transfer Protocol (HTTP) 502 Bad Gateway server error response code indicates that the server, while acting as a gateway or proxy, received an invalid response from the upstream server.

Since the proxy might be misbehaving, it can be seen as a temporary error.


mohsha-msft (approved):
👍

PR #209, opened 2020-09-10
[BlobSDK][GO] Feature: Set Blob Tags/Get Blob Tags/Find blobs by tags

mohsha-msft:


azblob/highlevel.go
```diff
@@ -133,7 +136,7 @@ func UploadBufferToBlockBlob(ctx context.Context, b []byte,
 		return nil, err
 	}
 	// All put blocks were successful, call Put Block List to finalize the blob
-	return blockBlobURL.CommitBlockList(ctx, blockIDList, o.BlobHTTPHeaders, o.Metadata, o.AccessConditions, o.BlobAccessTier)
+	return blockBlobURL.CommitBlockList(ctx, blockIDList, o.BlobHTTPHeaders, o.Metadata, o.AccessConditions, o.BlobAccessTier, nil)
```
kasobol-msft, line 139: should we pass tags here as well ?

azblob/url_append_blob.go
```diff
@@ -66,7 +67,7 @@ func (ab AppendBlobURL) Create(ctx context.Context, h BlobHTTPHeaders, metadata
 		ifModifiedSince, ifUnmodifiedSince, ifMatch, ifNoneMatch,
 		nil, // Blob tags
```
kasobol-msft, line 68: should this comment remain ?

azblob/url_blob.go
```diff
@@ -267,7 +308,7 @@ func (b BlobURL) StartCopyFromURL(ctx context.Context, source url.URL, metadata
 		nil, // Blob tags
```
kasobol-msft, line 308: should this comment remain? I can see many other places like that plz check them.

azblob/url_blob.go
```diff
@@ -72,6 +74,30 @@ func (b BlobURL) ToPageBlobURL() PageBlobURL {
...
 
+func SerializeBlobTagsHeader(blobTagsMap BlobTagsMap) *string {
+	if blobTagsMap == nil {
+		return nil
+	}
+	tags := make([]string, 0)
+	for key, val := range blobTagsMap {
+		tags = append(tags, url.QueryEscape(key)+"="+url.QueryEscape(val))
+	}
+	//tags = tags[:len(tags)-1]
```
rickle-msft, line 85: Delete

rickle-msft (approved):

azblob/highlevel.go
mohsha-msft: Thanks for pointing it out. Must have missed it.

azblob/url_append_blob.go
mohsha-msft: ifTags ( x-ms-if-tags conditional header to specify blob only if a specified condition is met) specify a SQL where clause on blob tags to operate only on blobs with a matching value. So wherever in function signature ifTags is mentioned, it facilitates the user to add a where clause. Since the primary task was to add set tags/get tags/find blob by tags, I did not add these changes. It can be rolled out separately. For reference, I looked at PR of other SDK ([.net](https://github.com/Azure/azure-sdk-for-net/pull/11060), [Python](https://github.com/Azure/azure-sdk-for-python/pull/11418))

kasobol-msft: thanks for confirming. The comment however is same for this param and for param changed few lines below and might be causing confusion.

mohsha-msft: Ohh I understand. I can do the changes but I do not wish to change it because 
1. These changes can be delivered separately with the current work. 
2. These changes were made when we added support for 2019-12-12 service version. If I do the changes, it also changes the git history of that line. 
What do you suggest?

kasobol-msft: I suggest to leave codebase in better shape and make these changes.

mohsha-msft: Done

kasobol-msft (approved):

PR #223, opened 2020-10-27
Writerat

nakulkar-msft:


guilhem:
Side effect of removing mmf:
we can now compile asure-storage-blob to wasm

PR #224, opened 2020-10-28
Set up CI with Azure Pipelines

zezha-msft:


azure-pipelines.yml
```diff
@@ -0,0 +1,28 @@
...
+- master
+- dev
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- task: GoTool@0
+  inputs:
+    version: '1.13'
```
mohsha-msft, line 11: version will be 1.15

mohsha-msft (changes requested):

PR #227, opened 2020-11-04
#7507887 #7507908 [Go][Blob] Support CPK-N/CPK-V

mohsha-msft:
Azure Storage automatically encrypts data (including metadata) stored in its cloud. Azure use AES256 scheme to encrypt/decrypt the data. Customer Provided Key (Name/Value) will facilitate request level encryption thereby giving granular control over encryption. 
For more info, please see [Encryption](https://docs.microsoft.com/en-us/azure/storage/common/storage-service-encryption)

Azure Storage supports following operations support sending customer-provided encryption keys on a request:
Put Blob, Put Block List, Put Block, Put Block from URL, Put Page, Put Page from URL, Append Block, Set Blob Properties, Set Blob Metadata, Get Blob, Get Blob Properties, Get Blob Metadata, Snapshot Blob.
For more info, please see [CPK](https://docs.microsoft.com/en-us/azure/storage/common/encryption-customer-provided-keys)

### DIFF
--- a/azblob/request_common.go
+++ b/azblob/request_common.go
@@ -6,11 +6,22 @@ package azblob
 // must be established to transfer the key.
 // Note: Azure Storage does not store or manage customer provided encryption keys. Keys are securely discarded
 // as soon as possible after they’ve been used to encrypt or decrypt the blob data.
+// https://docs.microsoft.com/en-us/azure/storage/common/storage-service-encryption
+// https://docs.microsoft.com/en-us/azure/storage/common/customer-managed-keys-overview
 type ClientProvidedKeyOptions struct {
-	EncryptionKey       *string                 // A Base64-encoded AES-256 encryption key value.
-	EncryptionKeySha256 *string                 // The Base64-encoded SHA256 of the encryption key.
-	EncryptionAlgorithm EncryptionAlgorithmType // Specifies the algorithm to use when encrypting data using the given key. Must be AES256.
-	EncryptionScope     *string
+	// A Base64-encoded AES-256 encryption key value.
+	EncryptionKey *string
+
+	// The Base64-encoded SHA256 of the encryption key.
+	EncryptionKeySha256 *string
+
+	// Specifies the algorithm to use when encrypting data using the given key. Must be AES256.
+	EncryptionAlgorithm EncryptionAlgorithmType
+
+	// Specifies the name of the encryption scope to use to encrypt the data provided in the request
+	// https://docs.microsoft.com/en-us/azure/storage/blobs/encryption-scope-overview
+	// https://docs.microsoft.com/en-us/azure/key-vault/general/overview
+	EncryptionScope *string
 }
 
 // NewClientProvidedKeyOptions function.
--- a/azblob/request_common.go
+++ b/azblob/request_common.go
@@ -1,0 +1,22 @@
+package azblob
+
+// ClientProvidedKeyOptions contains headers which may be be specified from service version 2019-02-02
+// or higher to encrypts the data on the service-side with the given key. Use of customer-provided keys
+// must be done over HTTPS. As the encryption key itself is provided in the request, a secure connection
+// must be established to transfer the key.
+// Note: Azure Storage does not store or manage customer provided encryption keys. Keys are securely discarded
+// as soon as possible after they’ve been used to encrypt or decrypt the blob data.
+type ClientProvidedKeyOptions struct {
+	EncryptionKey       *string                 // A Base64-encoded AES-256 encryption key value.
+	EncryptionKeySha256 *string                 // The Base64-encoded SHA256 of the encryption key.
+	EncryptionAlgorithm EncryptionAlgorithmType // Specifies the algorithm to use when encrypting data using the given key. Must be AES256.
+	EncryptionScope     *string
+}
+
+// NewClientProvidedKeyOptions function.
+// By default the value of encryption algorithm params is "AES256" for service version 2019-02-02 or higher.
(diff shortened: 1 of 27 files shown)


azblob/request_common.go
```diff
@@ -0,0 +1,22 @@
...
+// or higher to encrypts the data on the service-side with the given key. Use of customer-provided keys
+// must be done over HTTPS. As the encryption key itself is provided in the request, a secure connection
+// must be established to transfer the key.
+// Note: Azure Storage does not store or manage customer provided encryption keys. Keys are securely discarded
+// as soon as possible after they’ve been used to encrypt or decrypt the blob data.
+type ClientProvidedKeyOptions struct {
+	EncryptionKey       *string                 // A Base64-encoded AES-256 encryption key value.
+	EncryptionKeySha256 *string                 // The Base64-encoded SHA256 of the encryption key.
+	EncryptionAlgorithm EncryptionAlgorithmType // Specifies the algorithm to use when encrypting data using the given key. Must be AES256.
+	EncryptionScope     *string
```
gapra-msft, line 13: Could we add a short description for the encryption scope?

gapra-msft (commented):
The API Coverage looks good to me. The only thing that concerns me is that it may be confusing to have CPKN and CPKV both in the same ClientProvidedKeyOptions struct since they're technically different concepts

azblob/request_common.go
mohsha-msft: Yes I will.

mohsha-msft:
> The API Coverage looks good to me. The only thing that concerns me is that it may be confusing to have CPKN and CPKV both in the same ClientProvidedKeyOptions struct since they're technically different concepts

Since both CPK-N and CPK-V represents the same set of actions, I think it makes more sense to put it in the same wrapper together. However I understand your concern about confusion between CPK-N and CPK-V. For that I plan on 
1. Writing how-to Wiki on the Github page along with updates in documentation [here](https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#pkg-examples) so it shouldn't be a big problem then.
1. There are enough examples in the example files so this should resolve any confusion.

gapra-msft:
> > The API Coverage looks good to me. The only thing that concerns me is that it may be confusing to have CPKN and CPKV both in the same ClientProvidedKeyOptions struct since they're technically different concepts
> 
> Since both CPK-N and CPK-V represents the same set of actions, I think it makes more sense to put it in the same wrapper together. However I understand your concern about confusion between CPK-N and CPK-V. For that I plan on
> 
> 1. Writing how-to Wiki on the Github page along with updates in documentation [here](https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#pkg-examples) so it shouldn't be a big problem then.
> 2. There are enough examples in the example files so this should resolve any confusion.

Sounds good!

gapra-msft (approved):

azblob/zz_response_helpers.go
```diff
@@ -63,11 +63,9 @@ func (r *DownloadResponse) Body(o RetryReaderOptions) io.ReadCloser {
...
 	return NewRetryReader(r.ctx, r.Response(), r.getInfo, o,
 		func(ctx context.Context, getInfo HTTPGetterInfo) (*http.Response, error) {
-			resp, err := r.b.Download(ctx, getInfo.Offset, getInfo.Count,
-				BlobAccessConditions{
-					ModifiedAccessConditions: ModifiedAccessConditions{IfMatch: getInfo.ETag},
-				},
-				false)
+			resp, err := r.b.Download(ctx, getInfo.Offset, getInfo.Count, BlobAccessConditions{
+				ModifiedAccessConditions: ModifiedAccessConditions{IfMatch: getInfo.ETag},
+			}, false, ClientProvidedKeyOptions{})
```
zezha-msft, line 68: >ClientProvidedKeyOptions{} [](start = 13, length = 26)

What if the blob is encrypted? Don't we need to take the CPK info as an argument?

azblob/zt_examples_test.go
```diff
@@ -493,27 +493,24 @@ func ExampleBlobAccessConditions() {
...
 
 	// Create the blob (unconditionally; succeeds)
-	upload, err := blobURL.Upload(ctx, strings.NewReader("Text-1"), BlobHTTPHeaders{}, Metadata{}, BlobAccessConditions{}, DefaultAccessTier, nil)
+	upload, err := blobURL.Upload(ctx, strings.NewReader("Text-1"), BlobHTTPHeaders{}, Metadata{}, BlobAccessConditions{}, DefaultAccessTier, nil, ClientProvidedKeyOptions{})
 	showResult(upload, err)
 
 	// Download blob content if the blob has been modified since we uploaded it (fails):
-	showResult(blobURL.Download(ctx, 0, 0,
-		BlobAccessConditions{ModifiedAccessConditions: ModifiedAccessConditions{IfModifiedSince: upload.LastModified()}}, false))
+	showResult(blobURL.Download(ctx, 0, 0, BlobAccessConditions{ModifiedAccessConditions: ModifiedAccessConditions{IfModifiedSince: upload.LastModified()}}, false, ClientProvidedKeyOptions{}))
```
zezha-msft, line 500: >ClientProvidedKeyOptions{} [](start = 161, length = 26)

As we've discussed, the current design breaks cx needlessly when new parameters are introduced. This kind of feature should be easier to implement in track 2.

azblob/zz_response_helpers.go
mohsha-msft: Yes, that's correct. Thanks for pointing it out. This code snippet takes care of retrying download and would have failed with empty CPK. It was not failing earlier because all the blob downloads succeeded in the first attempt itself.

zezha-msft (approved):

Issue #233, opened 2020-11-27
Excessive memory use when uploading


ncw:
### Which version of the SDK was used?

v0.11.0

### Which platform are you using? (ex: Windows, Linux, Debian)

Linux

### What problem was encountered?

Excess memory usage compared to v0.8.0

### How can we reproduce the problem in the simplest way?

Uploading lots of files with rclone works pretty well.

### Have you found a mitigation/solution?

Setting `GOGC=20` mitigates the problem. This indicative of too much garbage being created.

### Analysis

In c0650f8c2c04b124963ffb252cdd21b8f1dd4074 #167 by @element-of-surprise the chunked uploader was refactored and a `sync.Pool` was introduced.

This `sync.Pool` appears to be causing memory problems according to the memory profiles below.

Testing with the azureblob backend in rclone produces the memory profile attached
[memprofile-1.53.3.log](https://github.com/Azure/azure-storage-blob-go/files/5608424/memprofile-1.53.3.log)

This looks like 
![profile002](https://user-images.githubusercontent.com/536803/100462334-0627ea80-30c2-11eb-9f08-49a77d7855ec.png)

Compared to the profile from the previous version of rclone, which shows no use of sync.Pool.

![profile001](https://user-images.githubusercontent.com/536803/100462415-2788d680-30c2-11eb-8f36-b3879ed75155.png)

This caused memory usage for rclone to go up from 400MB to 1.4 GB between the two versions. The old version of rclone was using v0.8.0

What appears to be happening is that memory is getting stuck in the `sync.Pool`.

From the [sync.Pool docs](https://golang.org/pkg/sync/#Pool):

>     On the other hand, a free list maintained as part of a short-lived object is
>     not a suitable use for a Pool, since the overhead does not amortize well in
>     that scenario. It is more efficient to have such objects implement their own
>     free list.

I think the use of sync.Pool is falling into that case. The Pool isn't shared between connections so it isn't really helping.

Because it is created and destroyed for each upload the old Pools will hang around in memory waiting to be garbage collected this is and the reason why GOGC=20 works around the problem I think.

Since you are recycling buffers of a fixed size maybe a `chan []byte` might work better instead of the `sync.Pool` - you could then flush it at the end of the transfer (unlike a `sync.Pool` which you can't flush).

Or alternatively implement a global `sync.Pool`. You'll need one for each size of chunk though which is a disadvantage.

[...]

PR #234, opened 2020-12-02
UploadStreamToBlockBlob gets TransferManager option

element-of-surprise:
See: https://github.com/Azure/azure-storage-blob-go/issues/233

There needs to be a more robust way of controling memory reuse and threadpooling for streaming copies (or maybe all copies, I will not speak to that here).

This introduces a new interface called TransferManager.  This provides 3 main fuctions:

- Control buffer allocation
- Control goroutine allocation
- Allow reuse

This allows a developer to control exactly how they wish to handle this for their use case.  We provide two implementations:
- Static buffer allocation similar to the original implementation
- Threadpool allocation similar to how it currently works since my change

Additionally, both of these support having shared pools between calls which is new.

It also allows for custom creation of implementations by users that better support their needs (a hybrid of the two above, one that adjusts buffer or concurrency based on transfer speed, ...)

Note: the behavior by default returns to the static buffer method here.  If azcopy likes the current behavior, it should switch using NewSyncPool()

Like the original change, this does not change the API compatibility.  I think that is probably something that should be done, but not in this change.

There are 3 files here that were not part of my change, but were not go fmt'd, so they have been updated by the go tool:
- bytes_wrtier.go
- zc_*

All tests without the end to end test system that are expected to pass do (comparing the failed tests in the non-changed version against this one).


element-of-surprise:
Please see conversation here:
https://github.com/Azure/azure-storage-blob-go/pull/242

azblob/highlevel.go
```diff
@@ -361,9 +362,148 @@ func DoBatchTransfer(ctx context.Context, o BatchTransferOptions) error {
...
+	if size < 1 || max < 1 {
+		return nil, fmt.Errorf("cannot be called with size or max set to < 1")
+	}
+
+	if size < _1MiB {
+		return nil, fmt.Errorf("cannot have size < 1MiB")
+	}
+
+	threadpool := make(chan func(), max)
+	buffers := make(chan []byte, max)
```
adreed-msft, line 402: Not sure how I feel about threadpooled anonymous functions having no access to the buffer which hosts them.

Also not sure how I feel about the threadpool itself being attached to the buffer, since it's not really used that way... How do you feel about separating these out?

adreed-msft:
I feel like in comparison to the other PR available, this one is much more user-friendly, which would be the vibe we're aiming for more.

adreed-msft (approved):

zezha-msft (approved):

PR #238, opened 2020-12-14
fix: chunkwriter hangs on cancel

davigust:
Chunkwriter `copyFromReader` will hang when canceling in response to a StageBlock error. `sendChunk` blocks on writing to the channel, but the writers exit on cancel without draining the channel. 

This change allows `sendChunk` to exit with an error in this scenario, rather than block on channel write.

Test added to repro.


mohsha-msft:
Running CI on https://dev.azure.com/azstorage/Go%20SDKs%20track%201.5/_build/results?buildId=3589&view=results

zezha-msft (approved):

PR #240, opened 2020-12-18
RetryPolicy: Retry for all network errors

nakulkar-msft:


azblob/zc_policy_retry.go
```diff
@@ -230,16 +230,8 @@ func NewRetryPolicyFactory(o RetryOptions) pipeline.Factory {
...
-						// Use non-retriable net.Error list, but not retriable list.
-						// As there are errors without Temporary() implementation,
-						// while need be retried, like 'connection reset by peer', 'transport connection broken' and etc.
-						// So the SDK do retry for most of the case, unless the error should not be retried for sure.
-						if !isNotRetriable(netErr) {
-							action = "Retry: net.Error and not in the non-retriable list"
-						} else {
-							action = "NoRetry: net.Error and in the non-retriable list"
-						}
+					} else if _, ok := err.(net.Error); ok {
```
mohsha-msft, line 233: Changes are fine but can you give some context on this?

azblob/zc_policy_retry.go
```diff
@@ -230,16 +230,8 @@ func NewRetryPolicyFactory(o RetryOptions) pipeline.Factory {
...
+						action = "Retry: net.Error"
```
zezha-msft, line 234: Some unretriable errors were identified in `isNotRetriable`, isn't this solution a bit of an overkill?

zezha-msft: In other words, we could perhaps improve `isNotRetriable` instead of not using it?

PR #242, opened 2020-12-18
chunkwriting: custom buffer pool

ajankovic:
Add option to `UploadStreamToBlockBlob` to specify custom pool for buffers.

This is needed to, for example, reuse pool between transfers.

Relates to #233 because with custom pool memory consumption remains constant.

### DIFF
--- a/azblob/chunkwriting.go
+++ b/azblob/chunkwriting.go
@@ -43,11 +43,16 @@ func copyFromReader(ctx context.Context, from io.Reader, to blockWriter, o Uploa
 		o:      o,
 		ch:     make(chan copierChunk, 1),
 		errCh:  make(chan error, 1),
-		buffers: sync.Pool{
+	}
+
+	if o.SyncPool != nil {
+		cp.buffers = o.SyncPool
+	} else {
+		cp.buffers = &sync.Pool{
 			New: func() interface{} {
 				return make([]byte, o.BufferSize)
 			},
-		},
+		}
 	}
 
 	// Starts the pools of concurrent writers.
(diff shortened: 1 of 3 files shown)


element-of-surprise:
Reviewers: Please see https://github.com/Azure/azure-storage-blob-go/pull/234 before allowing this in.  They solve the same issue but do it in different ways.  This here allows buffer pool reuse, the other allows complete customization (positives and negatives to them both).

ncw:
I like this fix. It is simple, it and directly addresses the problem of excess memory usage in the azure blob SDK.

When doing repeated transfers it will be necessary to set a `SyncPool` in order to stop memory usage blowing up in the SDK.

See discussion of the problem and the application of this fix here: https://forum.rclone.org/t/ask-for-settings-recommendation-for-azureblob/21505/19

element-of-surprise:
I considered doing this on my patch (and it was a really silly oversight on the original patch I put in that it didn't have this).

The question is the long haul.  With https://github.com/Azure/azure-storage-blob-go/pull/234, I decided that maybe the long term answer is to allow complete control by the user for the buffer AND the goroutines + reuse, with two reasonable defaults.

If a circular buffer cache is what is best for the use case, then you have it.  If you want the pool, you have it.  If you want a combination, you can do that (in certain use cases can really be efficient).  

You can also make versions that can respond to network congestion in services to optimize transfers (way beyond what I added).  

I'm actually 50/50 on this.  If the 99.9% use case is doing a few copies for some tool like rclone or azcopy, then it might not make sense and we should just use the sync.Pool as defined here.  It is beautifully simple.

If azblob think that we may have more complex cases that might want to optimize for themselves by using different buffering strategies and respond to buffer pressure from the network to optimize transfer rates, then something like 234 is probably the way to go.

adreed-msft (approved):
:shipit:

PR #246, opened 2021-01-15
Validating Client Request ID and Testing

siminsavani-msft:
The changes made in **_zc_policy_unique_request_id.go_** are made to validate that the client request id echoed back from the service matches. 

The new file **_zt_policy_request_id_test.go_** contains tests that handle different scenarios like:
1. In the case that Client Request ID is missing, the results from Do() are returned (**TestEchoClientRequestIDMissing**)
2. In the case that the error from Do() is not nil, the results from Do() are returned (**TestEchoClientRequestIDErrorFromNextPolicy**) 
3. In the case that Client Request ID is echoed back and does match, then results from Do() are returned (**TestEchoClientRequestIDMatch**)
4. In the case that Client Request ID is echoed back and does not match, then a error is returned (**TestEchoClientRequestIDNoMatch**)


azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+const (
+	// Testing scenarios for echoing Client Request ID
+	clientRequestIDMissing requestIDTestScenario = 1
+	clientRequestIDError   requestIDTestScenario = 2
+	clientRequestIDMatch   requestIDTestScenario = 3
+	clientRequestIDNoMatch requestIDTestScenario = 4
+)
+
+type dummyPolicy struct {
```
zezha-msft, line 22: Minor: to future-proof this struct, consider renaming it with key word "client request ID" to clarify that its intention is to mimic different behavior related to the client request ID.

Ex: if we add another dummy policy for a different purpose in the future, someone may have to come back here to rename it.

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (p dummyPolicy) Do(ctx context.Context, request pipeline.Request) (pipeline.Response, error) {
+	var header http.Header = make(map[string][]string)
+	var err error
+
+	// Set headers and errors according to each scenario
+	switch p.scenario {
+	case clientRequestIDMissing:
+		// header.Add("x-ms-client-request-id", "")
```
zezha-msft, line 34: Minor: remove code that is commented.

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+	case clientRequestIDError:
+		err = errors.New("error is not nil")
+	case clientRequestIDMatch:
+		header.Add("x-ms-client-request-id", p.matchID)
```
zezha-msft, line 38: For clientRequestIDMatch, couldn't you get the request ID from the argument `request`? Instead of using a variable on the struct.

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+	response := http.Response{Header: header}
+
+	return pipeline.NewHTTPResponse(&response), err
+}
+
+func (s *aztestsSuite) TestEchoClientRequestIDMissing(c *chk.C) {
+	factory := NewUniqueRequestIDPolicyFactory()
+
+	// Scenario 1: Client Request ID is missing
+	policy := factory.New(dummyPolicy{scenario: requestIDTestScenario(1)}, nil)
```
zezha-msft, line 54: Use the constant instead of requestIDTestScenario(1). Same for the other occurrences.

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (s *aztestsSuite) TestEchoClientRequestIDError(c *chk.C) {
+	factory := NewUniqueRequestIDPolicyFactory()
+
+	// Scenario 2: Do method returns an error
+	policy := factory.New(dummyPolicy{scenario: requestIDTestScenario(2)}, nil)
+	request, _ := pipeline.NewRequest("GET", url.URL{}, nil)
+	resp, err := policy.Do(context.Background(), request)
+
+	c.Assert(err, chk.NotNil)
```
zezha-msft, line 70: consider validating the actual err is what we expected.

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+func (s *aztestsSuite) TestEchoClientRequestIDMissing(c *chk.C) {
+	factory := NewUniqueRequestIDPolicyFactory()
+
+	// Scenario 1: Client Request ID is missing
+	policy := factory.New(dummyPolicy{scenario: requestIDTestScenario(1)}, nil)
+	request, _ := pipeline.NewRequest("GET", url.URL{}, nil)
+	resp, err := policy.Do(context.Background(), request)
+
+	c.Assert(err, chk.IsNil)
+	c.Assert(resp, chk.NotNil)
```
zezha-msft, line 59: Validate that the ID is indeed missing from `resp`. If validation steps are too generic, we can't be sure that the expectation is met.

In the future, consider practicing TDD (test driven development), it's a great way to write quality tests that guarantee good code. 💯

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Scenario 3: Client Request ID matches
+	matchRequestID := newUUID().String()
+	policy := factory.New(dummyPolicy{matchID: matchRequestID, scenario: requestIDTestScenario(3)}, nil)
+	request, _ := pipeline.NewRequest("GET", url.URL{}, nil)
+	request.Header.Set(xMsClientRequestID, matchRequestID)
+	resp, err := policy.Do(context.Background(), request)
+
+	c.Assert(err, chk.IsNil)
+	c.Assert(resp, chk.NotNil)
```
zezha-msft, line 85: validate that the resp ID matches.

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Scenario 4: Client Request ID does not match
+	matchRequestID := newUUID().String()
+	policy := factory.New(dummyPolicy{matchID: matchRequestID, scenario: requestIDTestScenario(4)}, nil)
+	request, _ := pipeline.NewRequest("GET", url.URL{}, nil)
+	request.Header.Set(xMsClientRequestID, matchRequestID)
+	resp, err := policy.Do(context.Background(), request)
+
+	c.Assert(err, chk.NotNil)
+	c.Assert(err.Error(), chk.Equals, "client Request ID from request and response does not match")
```
zezha-msft, line 99: Consider making the string "client Request ID from request and response does not match" into a constant, so that in case the message is updated we don't have to change the test.

azblob/zt_policy_request_id_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"net/http"
+	"net/url"
+)
+
+type requestIDTestScenario int
+
+const (
+	// Testing scenarios for echoing Client Request ID
+	clientRequestIDMissing requestIDTestScenario = 1
+	clientRequestIDError   requestIDTestScenario = 2
```
zezha-msft, line 17: The name `clientRequestIDError` is a bit confusing, since the error happens somewhere else, and not on the ID.

zezha-msft (changes requested):

azblob/zc_policy_unique_request_id.go
```diff
@@ -14,9 +14,20 @@ func NewUniqueRequestIDPolicyFactory() pipeline.Factory {
...
 			}
-			return next.Do(ctx, request)
+
+			resp, err := next.Do(ctx, request)
+
+			val := resp.Response().Header.Values(xMsClientRequestID)
+			if len(val) > 0 {
+				if val[0] != id {
+					err = errors.New("client Request ID from request and response does not match")
+				}
```
adreed-msft, line 29: I would prefer wrapping the old error in this one, if the old error isn't nil. [You should check out this news article on the Go website.](https://blog.golang.org/go1.13-errors)

adreed-msft (changes requested):
:clock1:

azblob/zc_policy_unique_request_id.go
zezha-msft: We don't have an error to wrap though, we should only check the client ID and throw the new error if there wasn't already an error.

zezha-msft (approved):

adreed-msft (approved):

PR #251, opened 2021-02-03
Adding Support for PutBlobFromUrl

siminsavani-msft:
Added support for PutBlobFromURL:

- Added _PutBlobFromURL_ in **url_block_blob.go** which calls PutBlobFromURL from regen code with appropriate parameters 
- Added tests for PutBlobFromURL in **zt_putblobfrom_url_test.go** including:

1. _TestPutBlockBlobFromURLWithTags_: checks if tags are copied over from source blob
2. _TestPutBlobFromURLWithSASReturnsVID_: checks if version ID is being returned in response header
3. _TestPutBlockBlobFromURL_: checks if block blob is copied to destination blob correctly
4. _TestSetTierOnPutBlockBlobFromURL_: checks if tiers are set correctly in destination blob
5. _TestPutBlobFromURLWithMissingSAS_: checks that PutBlobFromURL fails due to missing SAS
6. _TestPutBlobFromURLWithIncorrectURL_: checks that PutBlobFromURL fails due to incorrect URL

- Added fixes to regen code in **zt_blob_tags_test.go**, **zz_generated_models.go**, and **zt_url_container_test.go** 
- Commented out two tests (_TestUserDelegationSASContainer_ and _TestUserDelegationSASBlob_) which are failing due to service change.


azblob/url_block_blob.go
```diff
@@ -154,6 +154,22 @@ func (bb BlockBlobURL) CopyFromURL(ctx context.Context, source url.URL, metadata
...
 }
+
+// PutBlobFromURL synchronously creates a new Block Blob with data from the source URL up to a max length of 256MB.
+// For more information, see https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob-from-url.
+func (bb BlockBlobURL) PutBlobFromURL(ctx context.Context, h BlobHTTPHeaders, source url.URL, metadata Metadata, srcac ModifiedAccessConditions, dstac BlobAccessConditions, srcContentMD5 []byte, dstContentMD5 []byte, tier AccessTierType, blobTagsMap BlobTagsMap, cpk ClientProvidedKeyOptions) (*BlockBlobPutBlobFromURLResponse, error) {
+
+	srcIfModifiedSince, srcIfUnmodifiedSince, srcIfMatchETag, srcIfNoneMatchETag := srcac.pointers()
+	dstIfModifiedSince, dstIfUnmodifiedSince, dstIfMatchETag, dstIfNoneMatchETag := dstac.ModifiedAccessConditions.pointers()
+	dstLeaseID := dstac.LeaseAccessConditions.pointers()
+	blobTagsString := SerializeBlobTagsHeader(blobTagsMap)
```
adreed-msft, line 167: Code cleanliness: If the order of .pointers() matches a subsection of the function, you can place .pointers() in there and move along.

azblob/highlevel.go
```diff
@@ -185,7 +185,7 @@ type DownloadFromBlobOptions struct {
 
 // downloadBlobToWriterAt downloads an Azure blob to a buffer with parallel.
 func downloadBlobToWriterAt(ctx context.Context, blobURL BlobURL, offset int64, count int64,
-	writer io.WriterAt, o DownloadFromBlobOptions, initialDownloadResponse *DownloadResponse) error {
+	writer io.WriterAt, o DownloadFromBlobOptions, initialDownloadResponse *downloadResponse) error {
```
zezha-msft, line 188: >downloadResponse [](start = 73, length = 16)

This looks like an accidental edit from the refactoring. It should stay  DownloadResponse

zezha-msft:
>    "properties": [

Not sure what this is but there's no need to check in this file

---
Refers to: swagger/Go_BlobStorage/code-model-v1:6 in 5787ac4. [](commit_id = 5787ac453d9adc8ddebe01c55bcbc3f955888021, deletion_comment = False)

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,302 @@
...
+	container, _ := createNewContainer(c, bsu)
+	defer delContainer(c, container)
+
+	testSize := 1 * 1024 * 1024 // 1MB
+	r, sourceData := getRandomDataAndReader(testSize)
+	sourceDataMD5Value := md5.Sum(sourceData)
+	srcBlob := container.NewBlockBlobURL("srcBlob")
+	destBlob := container.NewBlockBlobURL("destBlob")
+
+	blobTagsMap := BlobTagsMap{
```
zezha-msft, line 27: >blobTagsMap [](start = 1, length = 11)

the tags don't seem to be checked on the result blob

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,302 @@
...
+	c.Assert(err, chk.IsNil)
+	c.Assert(destData, chk.DeepEquals, sourceData)
+	c.Assert(len(downloadResp.NewMetadata()), chk.Equals, 1)
+
+	// Edge case 1: Provide bad MD5 and make sure the put fails
+	_, badMD5 := getRandomDataAndReader(16)
+	_, err = destBlob.PutBlobFromURL(ctx, BlobHTTPHeaders{}, srcBlobURLWithSAS, Metadata{}, ModifiedAccessConditions{}, BlobAccessConditions{}, badMD5, badMD5, DefaultAccessTier, blobTagsMap, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.NotNil)
+
+	// Edge case 2: Not providing any source MD5 should see the CRC getting returned instead
```
zezha-msft, line 76: >// Edge case 2: Not providing any source MD5 should see the CRC getting returned instead [](start = 1, length = 88)

the test did not check CRC though?

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,302 @@
...
+	c.Assert(resp.Date().IsZero(), chk.Equals, false)
+	c.Assert(resp.ContentMD5(), chk.DeepEquals, sourceDataMD5Value[:])
+
+	// Check data integrity through downloading.
+	downloadResp, err := destBlob.BlobURL.Download(ctx, 0, CountToEnd, BlobAccessConditions{}, false, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.IsNil)
+	destData, err := ioutil.ReadAll(downloadResp.Body(RetryReaderOptions{}))
+	c.Assert(err, chk.IsNil)
+	c.Assert(destData, chk.DeepEquals, sourceData)
+	c.Assert(len(downloadResp.NewMetadata()), chk.Equals, 1)
```
zezha-msft, line 69: >c.Assert(len(downloadResp.NewMetadata()), chk.Equals, 1 [](start = 1, length = 55)

test should check all parameters explicitly, for example we can check the actual value of metadata

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,302 @@
...
+	srcBlobParts.SAS, err = BlobSASSignatureValues{
+		Protocol:      SASProtocolHTTPS,                     // Users MUST use HTTPS (not HTTP)
+		ExpiryTime:    time.Now().UTC().Add(48 * time.Hour), // 48-hours before expiration
+		ContainerName: srcBlobParts.ContainerName,
+		BlobName:      srcBlobParts.BlobName,
+		Permissions:   BlobSASPermissions{Read: true}.String(),
+	}.NewSASQueryParameters(credential)
+	if err != nil {
+		c.Fatal(err)
+	}
```
zezha-msft, line 115: The source set up code could perhaps be extracted as a method to reduce test length.

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,302 @@
...
+	_, badMD5 := getRandomDataAndReader(16)
+	_, err = destBlob.PutBlobFromURL(ctx, BlobHTTPHeaders{}, srcBlobURLWithSAS, Metadata{}, ModifiedAccessConditions{}, BlobAccessConditions{}, badMD5, badMD5, DefaultAccessTier, BlobTagsMap{}, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.NotNil)
+
+	// Edge case 2: Not providing any source MD5 should see the CRC getting returned instead
+	resp, err = destBlob.PutBlobFromURL(ctx, BlobHTTPHeaders{}, srcBlobURLWithSAS, Metadata{}, ModifiedAccessConditions{}, BlobAccessConditions{}, nil, nil, DefaultAccessTier, nil, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.IsNil)
+	c.Assert(resp.Response().StatusCode, chk.Equals, 201)
+	c.Assert(resp.Response().Header.Get("x-ms-version"), chk.Equals, ServiceVersion)
+	c.Assert(resp.Response().Header.Get("x-ms-version-id"), chk.NotNil)
```
zezha-msft, line 145: the edge cases seem to be repeated across multiple tests, is this necessary?

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,302 @@
...
+	ctx := context.Background() // Use default Background context
+	srcBlob := container.NewBlockBlobURL(generateBlobName())
+	destBlob := container.NewBlockBlobURL(generateBlobName())
+
+	// Prepare source blob for put.
+	uploadSrcResp, err := srcBlob.Upload(ctx, r, BlobHTTPHeaders{}, Metadata{}, BlobAccessConditions{}, DefaultAccessTier, nil, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.IsNil)
+	c.Assert(uploadSrcResp.Response().StatusCode, chk.Equals, 201)
+
+	// Invoke put blob from URL with URL without SAS
```
zezha-msft, line 281: >// Invoke put blob from URL with URL without SAS [](start = 1, length = 48)

perhaps validate that the error is what you expected

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,302 @@
...
+	container, _ := createNewContainer(c, bsu)
+	defer delContainer(c, container)
+
+	testSize := 8 * 1024 * 1024 // 8MB
+	_, sourceData := getRandomDataAndReader(testSize)
+	sourceDataMD5Value := md5.Sum(sourceData)
+	ctx := context.Background() // Use default Background context
+	destBlob := container.NewBlockBlobURL(generateBlobName())
+
+	// Invoke put blob from URL with incorrect URL
```
zezha-msft, line 298: >// Invoke put blob from URL with incorrect URL [](start = 1, length = 46)

perhaps validate that the error is what you expected

zezha-msft (changes requested):
:clock1:

mohsha-msft:
I think the branch you're requesting to merge into ... should be `dev` instead of  `master` . We merge changes from `dev` into `master` during release.

azblob/zt_put_blob_from_url_test.go
siminsavani-msft: They are repetitive for the most part so I kept the edge cases on the first test method. Also, for this test, the version ID is being checked in the edge case so I kept this one.

azblob/url_block_blob.go
siminsavani-msft: Tried doing this but Golang doesn't seem to recognize that the '.pointers()' returns the desired subsection and thinks it is only filling one parameter.

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,246 @@
...
+	c.Assert(resp.ContentMD5(), chk.DeepEquals, sourceDataMD5Value[:])
+
+	// Check data integrity through downloading.
+	downloadResp, err := destBlob.BlobURL.Download(ctx, 0, CountToEnd, BlobAccessConditions{}, false, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.IsNil)
+	destData, err := ioutil.ReadAll(downloadResp.Body(RetryReaderOptions{}))
+	c.Assert(err, chk.IsNil)
+	c.Assert(destData, chk.DeepEquals, sourceData)
+	c.Assert(len(downloadResp.NewMetadata()), chk.Equals, 1)
+	c.Assert(downloadResp.r.rawResponse.Header.Get("x-ms-meta-foo"), chk.Equals, "bar")
```
zezha-msft, line 76: >c.Assert(downloadResp.r.rawResponse.Header.Get("x-ms-meta-foo"), chk.Equals, "bar") [](start = 1, length = 83)

You can check metadata in a much easier way, no need to go through the raw response. Please refer to the other tests.

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,246 @@
...
+	// Edge case: Not providing any source MD5 should see the CRC getting returned instead and service version matches
+	resp, err = destBlob.PutBlobFromURL(ctx, BlobHTTPHeaders{}, srcBlobURLWithSAS, Metadata{}, ModifiedAccessConditions{}, BlobAccessConditions{}, nil, nil, DefaultAccessTier, nil, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.IsNil)
+	c.Assert(resp.Response().StatusCode, chk.Equals, 201)
+	c.Assert(resp.rawResponse.Header.Get("x-mx-content-crc64"), chk.NotNil)
+	c.Assert(resp.Response().Header.Get("x-ms-version"), chk.Equals, ServiceVersion)
+	c.Assert(resp.Response().Header.Get("x-ms-version-id"), chk.NotNil)
+}
+
+func (s *aztestsSuite) TestPutBlockBlobFromURL(c *chk.C) {
```
zezha-msft, line 141: Minor: put simplest case first.

azblob/zt_put_blob_from_url_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// Check data integrity through downloading.
+	downloadResp, err := destBlob.BlobURL.Download(ctx, 0, CountToEnd, BlobAccessConditions{}, false, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.IsNil)
+	destData, err := ioutil.ReadAll(downloadResp.Body(RetryReaderOptions{}))
+	c.Assert(err, chk.IsNil)
+	c.Assert(destData, chk.DeepEquals, sourceData)
+
+	// Make sure the metadata got copied over
+	c.Assert(len(downloadResp.NewMetadata()), chk.Equals, 1)
```
zezha-msft, line 184: >c.Assert(len(downloadResp.NewMetadata()), chk.Equals, 1) [](start = 1, length = 56)

please assert actual value.

zezha-msft (approved):
The file `swagger/Go_BlobStorage/code-model-v1` is still there, please make sure to remove it before merging.

adreed-msft (approved):
:shipit:

PR #252, opened 2021-02-03
Removed requestId, Timeout, versionId, snapshot from the function signature of GetTags() and SetTags()

mohsha-msft:


azblob/url_blob.go
```diff
@@ -146,15 +164,18 @@ func (b BlobURL) Delete(ctx context.Context, deleteOptions DeleteSnapshotsOption
...
 }
 
 // GetTags operation enables users to get tags on a blob or specific blob version, or snapshot.
 // https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob-tags
-func (b BlobURL) GetTags(ctx context.Context, timeout *int32, requestID *string, snapshot *string, versionID *string, ifTags *string) (*BlobTags, error) {
-	return b.blobClient.GetTags(ctx, timeout, requestID, snapshot, versionID, ifTags)
+func (b BlobURL) GetTags(ctx context.Context, ifTags *string) (*BlobTags, error) {
+	versionID := b.getVersionId()
+	snapshot := b.getSnapshot()
+	return b.blobClient.GetTags(ctx, nil, nil, snapshot, versionID, ifTags)
```
zezha-msft, line 178: I'm confused why versionID and snaptshotID need to be passed this way for this API.

Every other call to the protocol layer passes `nil`, and rely on the URL which already contains this info, right?

mohsha-msft: Hey Ze, I've already added that as a part of Blob URL and inside the `GetTags()` we're fetching it from the URL.

We're doing the same thing in Python SDK [here](https://github.com/Azure/azure-sdk-for-python/blob/e2dba75181f8b4336ae57e75aa391322c12c3123/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py#L2596). The function signature is similar to what we're using right now.

mohsha-msft: I wonder why they have versionId and snapshotID in the function signature then.

zezha-msft (approved):
:shipit:

PR #264, opened 2021-03-10
Keep compatibility with Go 1.13.14.

semk:
`http.Header` has no method `Values` on older go versions. azblob 0.13 broke the compatibility with Go 1.13.

### DIFF
--- a/azblob/zc_policy_unique_request_id.go
+++ b/azblob/zc_policy_unique_request_id.go
@@ -22,7 +22,8 @@ func NewUniqueRequestIDPolicyFactory() pipeline.Factory {
 			resp, err := next.Do(ctx, request)
 
 			if err == nil && resp != nil {
-				if resp.Response().Header.Get(xMsClientRequestID) != id {
+				crId := resp.Response().Header.Get(xMsClientRequestID)
+				if crId != "" && crId != id {
 					err = errors.New("client Request ID from request and response does not match")
 				}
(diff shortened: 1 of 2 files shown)


mohsha-msft (approved):
Changes look good to me. I remember these changes were made while upgrading the Go version to 1.15. To keep it compatible with Go version 1.13.x, we should do those changes.

semk:
@mohsha-msft Updated the changes to not return the error when the request id is missing (as err was not expected by the tests). How can I run the tests locally?

mohsha-msft:
@semk ,

Thanks for reaching out and I appreciate your contribution.
As for your query about running the test suite locally, you need to setup these environment variables
```
ACCOUNT_NAME
ACCOUNT_KEY
BLOB_STORAGE_ACCOUNT_NAME
BLOB_STORAGE_ACCOUNT_KEY
PREMIUM_ACCOUNT_NAME
PREMIUM_ACCOUNT_KEY
SECONDARY_ACCOUNT_NAME
SECONDARY_ACCOUNT_KEY
APPLICATION_ID
CLIENT_SECRET
TENANT_ID
```

Most of the test cases will work with just the first two (`ACCOUNT_NAME` and `ACCOUNT_KEY`)
Please specify them and you can run the tests locally.

Reach out for more help :)

siminsavani-msft (approved):

PR #269, opened 2021-03-24
Reverting changes to GetTags/SetTags & BlobPropertiesInternal

siminsavani-msft:
Reverting changes made to GetTags/SetTags in PutBlobFromUrl PR and changing BlobPropertiesInternal to BlobProperties to avoid breaking changes in AzCopy.


mohsha-msft (commented):
LGTM.

Issue #260, opened 2021-03-02
SDK Panics When Concurrently Uploading 10 Large Blobs


marwan-at-work:
### Which version of the SDK was used?
v0.13.0
### Which platform are you using? (ex: Windows, Linux, Debian)
Darwin and Linux
### What problem was encountered?
Trying to concurrently upload 10 large files to the same container results in the following panic: 

```
uploading file of size 393472124
uploading file of size 347726145
uploading file of size 349536845
uploading file of size 355056046
uploading file of size 347411174
uploading file of size 355576358
uploading file of size 393468713
uploading file of size 393473126
uploading file of size 306350950
uploading file of size 349612217

panic: send on closed channel

goroutine 279 [running]:
github.com/Azure/azure-storage-blob-go/azblob.staticBuffer.Put(0xc000502a20, 0x800000, 0xc0005029c0, 0xc00bb00000, 0x800000, 0x800000)
        /Users/home/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/highlevel.go:427 +0x53
github.com/Azure/azure-storage-blob-go/azblob.(*copier).write(0xc0000c01a0, 0xc00bb00000, 0x800000, 0x800000, 0xc01ac53200, 0x58)
        /Users/home/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/chunkwriting.go:166 +0x413
github.com/Azure/azure-storage-blob-go/azblob.(*copier).sendChunk.func1()
        /Users/home/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/chunkwriting.go:136 +0xc5
github.com/Azure/azure-storage-blob-go/azblob.NewStaticBuffer.func1(0xc0005029c0)
        /Users/home/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/highlevel.go:406 +0x3b
created by github.com/Azure/azure-storage-blob-go/azblob.NewStaticBuffer
        /Users/home/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/highlevel.go:404 +0xe5
exit status 2
```
### How can we reproduce the problem in the simplest way?

Grab 10 large files (300-400 megabytes) and run the following code concurrently: 

```golang
import (
	"gocloud.dev/blob"
)

func copy(ctx context, key string, azbucket *blob.Bucket) error {
[...]

Issue #257, opened 2021-02-17
Panic when calling UploadStreamToBlockBlob with invalid credentials


guy-har:
### Which version of the SDK was used?
0.13.0
### Which platform are you using? (ex: Windows, Linux, Debian)
Mac
### What problem was encountered?
Calling `UploadStreamToBlockBlob`  with wrong credentials throws panic
The reason for that is that is that `copyFromReader` returns without waiting for goroutines to end ( In case of error) 
### How can we reproduce the problem in the simplest way?
I ran into this bug when I  Called `UploadStreamToBlockBlob` with a ~11MB size file with invalid credentials.
However a simple way would be to run this test 
```
func Test_CopyWithErr(t *testing.T) {
	ctx := context.Background()
	p, err := createSrcFile(_1MiB * 12)
	if err != nil {
		panic(err)
	}
	defer os.Remove(p)

	from, err := os.Open(p)
	if err != nil {
		panic(err)
	}

	br := newFakeBlockWriter()
	defer br.cleanup()

	br.errOnBlock = 1
	transferManager, err := NewStaticBuffer(_1MiB, 1)
	if err != nil {
		t.Fatal(err)
	}
	defer transferManager.Close()
	_, err = copyFromReader(ctx, from, br, UploadStreamToBlockBlobOptions{TransferManager: transferManager})
	
}
```
### Have you found a mitigation/solution?
yes, calling wait() before returning the error

siminsavani-msft:
@guy-har Hi! This issue should be resolved with the new version!

PR #275, opened 2021-04-22
Resolving Concurrency Issues 

siminsavani-msft:
Resolution for these github issues: https://github.com/Azure/azure-storage-blob-go/issues/260, https://github.com/Azure/azure-storage-blob-go/issues/257

Will add additional tests and comments!

### DIFF
--- a/azblob/zt_highlevel_test.go
+++ b/azblob/zt_highlevel_test.go
@@ -432,3 +432,31 @@ func (s *aztestsSuite) TestDoBatchTransferWithError(c *chk.C) {
 	mmf.isClosed = true
 	time.Sleep(time.Second * 5)
 }
+
+func (s *aztestsSuite) Test_CopyFromReader(c *chk.C) {
+	ctx := context.Background()
+	p, err := createSrcFile(_1MiB * 12)
+	if err != nil {
+		panic(err)
+	}
+
+	defer os.Remove(p)
+
+	from, err := os.Open(p)
+	if err != nil {
+		panic(err)
+	}
+
+	br := newFakeBlockWriter()
+	defer br.cleanup()
+
+	br.errOnBlock = 1
+	transferManager, err := NewStaticBuffer(_1MiB, 1)
+	if err != nil {
+		panic(err)
+	}
+	defer transferManager.Close()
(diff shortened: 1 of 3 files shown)


azblob/zt_highlevel_test.go
```diff
@@ -432,3 +432,31 @@ func (s *aztestsSuite) TestDoBatchTransferWithError(c *chk.C) {
 	mmf.isClosed = true
 	time.Sleep(time.Second * 5)
 }
+
+func (s *aztestsSuite) Test_CopyFromReader(c *chk.C) {
+	ctx := context.Background()
+	p, err := createSrcFile(_1MiB * 12)
+	if err != nil {
+		panic(err)
```
zezha-msft, line 440: Assert instead of panic.

zezha-msft (approved):

azblob/zt_highlevel_test.go
```diff
@@ -432,3 +432,31 @@ func (s *aztestsSuite) TestDoBatchTransferWithError(c *chk.C) {
...
+	}
+
+	br := newFakeBlockWriter()
+	defer br.cleanup()
+
+	br.errOnBlock = 1
+	transferManager, err := NewStaticBuffer(_1MiB, 1)
+	if err != nil {
+		panic(err)
```
mohsha-msft, line 456: c.assert(err, chk.NotNil)

azblob/zt_highlevel_test.go
```diff
@@ -432,3 +432,31 @@ func (s *aztestsSuite) TestDoBatchTransferWithError(c *chk.C) {
 	mmf.isClosed = true
 	time.Sleep(time.Second * 5)
 }
+
+func (s *aztestsSuite) Test_CopyFromReader(c *chk.C) {
+	ctx := context.Background()
+	p, err := createSrcFile(_1MiB * 12)
```
mohsha-msft, line 438: Just Curious: Why `12MiB`? If this any random number, can we keep the size a little less as it increases testsuite runtime on local

mohsha-msft (approved):

azblob/zt_highlevel_test.go
siminsavani-msft: So I was hoping to test a larger file, but we could potentially keep it smaller.

adreed-msft (approved):

PR #281, opened 2021-07-01
v0.14.0 Release

siminsavani-msft:
- Updated Get Blob Tags and Set Blob Tags function signatures
- Added Put Blob From URL
- Offer knob to disable application logging (Syslog)
- Added examples for MSI Login
- Updated go.mod to address dependency issues
- Fixed issues [#260](https://github.com/Azure/azure-storage-blob-go/issues/260) and [#257](https://github.com/Azure/azure-storage-blob-go/issues/257)


zezha-msft (approved):
Looks good! Btw the version is in this file: https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/version.go

Please update it to 0.14 accordingly.

zezha-msft:
There seems to be a flaky test, I triggered a rerun

PR #282, opened 2021-07-04
Avoiding unecessary cast (linitng)

meox:


zezha-msft:
@meox thanks for your contribution!

PR #284, opened 2021-07-09
[chunkwriting] Return original buffer to the pool

kfarnung:
When the buffer isn't filled to capacity a new slice is created that's
smaller than the original. In that case the smaller slice is returned to
the pool which prevents the rest of the capacity from being used again.

The solution is to pass the original slice through and attach the
length. This allows the original slice to be returned to the pool once
the operation is complete.

This change also simplifies the `sendChunk` method, ensuring that the
buffer is returned to the TransferManager even when no bytes were read
from the reader.


serprex:
It looks like this PR should also be applied to https://github.com/Azure/azure-sdk-for-go

https://github.com/Azure/azure-sdk-for-go/blob/dfdc424c554cfb6e45a53e259c5c215bb4e1d3a6/sdk/storage/azblob/chunkwriting.go#L112

mohsha-msft (approved):
Ohh I get what you're trying to do. Thanks for the contribution!

PR #291, opened 2021-07-23
Adding BlobDeleteType to Delete API

siminsavani-msft:


azblob/zt_url_service_test.go
```diff
@@ -206,6 +206,78 @@ func (s *aztestsSuite) TestAccountListContainersMaxResultsSufficient(c *chk.C) {
 	c.Assert(len(response.ContainerItems) >= 2, chk.Equals, true)
 }
 
+func (s *aztestsSuite) TestPermanentDeleteAndUndelete(c *chk.C) {
+	c.Skip("Versioning on storage account must be disabled in order for soft delete to function.")
```
mohsha-msft, line 210: Instead of skipping, try using secondary account to run the test cases.
https://github.com/Azure/azure-storage-blob-go/blob/da3335192ee9609890f4172c43d0750c65941b6f/azblob/zt_test.go#L340

```go
func getAlternateBSU() (ServiceURL, error) {
	return getGenericBSU("SECONDARY_")
}

func getPremiumBSU() (ServiceURL, error) {
	return getGenericBSU("PREMIUM_")
}

func getBlobStorageBSU() (ServiceURL, error) {
	return getGenericBSU("BLOB_STORAGE_")
}
```

mohsha-msft: Tests are important for this particular scenario.

mohsha-msft (changes requested):

azblob/url_blob.go
```diff
@@ -143,6 +143,16 @@ func (b BlobURL) Delete(ctx context.Context, deleteOptions DeleteSnapshotsOption
 		nil, BlobDeleteNone)
 }
 
+// PermanentDelete permanently deletes-soft deleted snapshots & soft-deleted version blobs and is a dangerous operation and SHOULD NOT BE USED.
```
zezha-msft, line 146: typo: the dash is placed in the wrong place

zezha-msft: Should not be used unless you know exactly the implications. We will not provide support for this API.

azblob/url_blob.go
```diff
@@ -143,6 +143,16 @@ func (b BlobURL) Delete(ctx context.Context, deleteOptions DeleteSnapshotsOption
...
+// For more information, see https://docs.microsoft.com/rest/api/storageservices/delete-blob.
+func (b BlobURL) PermanentDelete(ctx context.Context, deleteOptions DeleteSnapshotsOptionType, ac BlobAccessConditions, blobDeleteType BlobDeleteType) (*BlobDeleteResponse, error) {
+	ifModifiedSince, ifUnmodifiedSince, ifMatchETag, ifNoneMatchETag := ac.ModifiedAccessConditions.pointers()
+	return b.blobClient.Delete(ctx, nil, nil, nil, ac.LeaseAccessConditions.pointers(), deleteOptions,
+		ifModifiedSince, ifUnmodifiedSince, ifMatchETag, ifNoneMatchETag,
+		nil, // Blob ifTags
+		nil, blobDeleteType)
```
zezha-msft, line 153: shouldn't the delete type be set already? Since the function is already called permanent delete

azblob/zt_url_service_test.go
```diff
@@ -206,6 +206,80 @@ func (s *aztestsSuite) TestAccountListContainersMaxResultsSufficient(c *chk.C) {
...
+	for i := 0; i < 2; i++ {
+		testSize := 8 * 1024
+		r, _ := getRandomDataAndReader(testSize)
+		blob, _ := getBlockBlobURL(c, containerURL)
+		blobURLs = append(blobURLs, blob)
+		cResp, err := blob.Upload(ctx, r, BlobHTTPHeaders{}, nil, BlobAccessConditions{}, DefaultAccessTier, nil, ClientProvidedKeyOptions{})
+
+		c.Assert(err, chk.IsNil)
+		c.Assert(cResp.StatusCode(), chk.Equals, 201)
+		time.Sleep(time.Second * 30)
```
zezha-msft, line 237: no need to sleep for 30 sec here right?

azblob/zt_url_service_test.go
```diff
@@ -206,6 +206,80 @@ func (s *aztestsSuite) TestAccountListContainersMaxResultsSufficient(c *chk.C) {
...
+	// Soft delete first blob
+	delResp, err := blobURLs[0].Delete(ctx, DeleteSnapshotsOptionNone, BlobAccessConditions{}) //soft delete
+	c.Assert(err, chk.IsNil)
+	c.Assert(delResp, chk.NotNil)
+	time.Sleep(time.Second * 30)
+
+	// Undelete soft deleted blob
+	undelResp, err := blobURLs[0].Undelete(ctx)
+	c.Assert(err, chk.IsNil)
+	c.Assert(undelResp, chk.NotNil)
```
zezha-msft, line 249: validate status code directly

azblob/zt_url_service_test.go
```diff
@@ -206,6 +206,80 @@ func (s *aztestsSuite) TestAccountListContainersMaxResultsSufficient(c *chk.C) {
...
+	c.Assert(delResp, chk.NotNil)
+
+	// Permanent delete snapshot
+	delResp, err = snapshotBlob.PermanentDelete(ctx, DeleteSnapshotsOptionNone, BlobAccessConditions{}, BlobDeletePermanent)
+	c.Assert(err, chk.IsNil)
+	c.Assert(delResp, chk.NotNil)
+
+	// Check that snapshot has been deleted
+	_, err = snapshotBlob.GetProperties(ctx, BlobAccessConditions{}, ClientProvidedKeyOptions{})
+	c.Assert(err, chk.NotNil)
```
zezha-msft, line 275: validate more directly by looking at status code

azblob/zt_url_service_test.go
```diff
@@ -206,6 +206,80 @@ func (s *aztestsSuite) TestAccountListContainersMaxResultsSufficient(c *chk.C) {
...
+
+	// Check that both blobs exist
+	listBlobResp1, err = containerURL.ListBlobsFlatSegment(ctx, Marker{}, ListBlobsSegmentOptions{Details: BlobListingDetails{Deleted: true, Snapshots: true}})
```
zezha-msft, line 278: how was the blob at [0] useful? It seems that it forces the listing validation to be more generic than necessary

zezha-msft (approved):

azblob/zt_url_service_test.go
siminsavani-msft: I used it to test if undelete works earlier.

PR #295, opened 2021-08-18
[azblob] Fix issue with context timeout below 1 second

bmermet:
The current logic truncates timeouts to the second, this means that any context with less than a second remaining before the deadline systematically fails. This PR switches to using time.Time instead of integers to manipulate the timeouts to address this. 

I have tested this change locally, but I'm happy to add a unit test if someone can point me towards the test file.

### DIFF
--- a/azblob/zc_policy_retry.go
+++ b/azblob/zc_policy_retry.go
@@ -181,25 +181,25 @@ func NewRetryPolicyFactory(o RetryOptions) pipeline.Factory {
 				}
 
 				// Set the server-side timeout query parameter "timeout=[seconds]"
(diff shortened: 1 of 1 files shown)


siminsavani-msft:
Hi @bmermet! Thank you for your contribution. You can add tests in this file: https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/zt_policy_retry_test.go.

bmermet:
Hi @siminsavani-msft, sorry for the lack of reactivity, I was on PTO the past two week.
Thank you for reviewing and merging this.

PR #331, opened 2023-02-21
Redact CPK headers in log.

mkrautz:
This ensures that CPK-related headers (x-ms-encryption-key, x-ms-encryption-key-sha256) are redacted in log output, ensuring that key material doesn't leak.


mkrautz:
This is part of a change in azcopy that ensures CPK headers are redacted in log output:
https://github.com/Azure/azure-storage-azcopy/pull/2080

